PR #4, opened 2021-01-10
JOIN command

epoberezkin:
It compiles but it is not working yet (store throws exception).

I won't to add test harness before debugging this issue - it'll be another PR.


epoberezkin:
It probably misses updateQueueStatus implementation

efim-poberezkin (approved):

PR #6, opened 2021-01-11
agent store: add updateQueueStatus tests

efim-poberezkin:


tests/AgentTests/SQLite.hs
```diff
@@ -34,11 +34,13 @@ storeTests = withStore do
...
     describe "addRcvQueue" testAddRcvQueue
-    describe "deleteConnReceive" do
+    describe "deleteConn" do
       describe "Receive connection" testDeleteConnReceive
       describe "Send connection" testDeleteConnSend
-
--- describe "deleteConnDuplex" testDeleteConnDuplex
+      describe "Duplex connection" testDeleteConnDuplex
+    describe "updateQueueStatus" testUpdateQueueStatus
+    describe "updateQueueStatusBadDirectionSnd" testUpdateQueueStatusBadDirectionSnd
```
epoberezkin, line 42: Better to group under the method, so the title is either a method name or plain English

epoberezkin (changes requested):
One comment re test names/grouping otherwise ok.

Probably good to also add successful change status test for Send/Recive connections - in the next PR

epoberezkin (approved):

PR #12, opened 2021-01-12
ci: rename workflow and job

efim-poberezkin:


efim-poberezkin:
Experimentation with build badge

PR #14, opened 2021-01-13
agent store: add createMsg

efim-poberezkin:


src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -94,6 +94,7 @@ data DeliveryStatus
 
 type SMPServerId = Int64
 
+-- TODO rework types - decouple Transmission types from Store? Convert on the agent instead?
 class Monad m => MonadAgentStore s m where
```
epoberezkin, line 98: yes, I agree, we need to extract all types (and, possibly, some functions) used across Agent, Server and Client into a separate top level file (e.g. Core - tbc), so that  Server and Agent do not depend on each other.

There is no rush doing it - I think it's best to do it in the end once we can see what depends on what and decide how we want the dependency graph to look.

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -412,15 +429,27 @@ instance (MonadUnliftIO m, MonadError StoreError m) => MonadAgentStore SQLiteSto
         (rcvQId, _) <- getConnection st connAlias
         case rcvQId of
           Just qId -> updateRcvQueueStatus st qId status
-          Nothing -> throwError SEBadConn
+          Nothing -> throwError SEBadQueueDirection
       SND -> do
```
epoberezkin, line 433: That's another thing I don't like - it looks like a command, we could rename for example to DSend/DirSend/DirectionSend. No need to change in this PR

src/Simplex/Messaging/Agent/Transmission.hs
```diff
@@ -91,13 +90,50 @@ data ACommand (p :: AParty) where
...
   REPLY :: SMPQueueInfo -> AMessage
   A_MSG :: MsgBody -> AMessage
 
--- A_ACK :: AgentMsgId -> AckStatus -> AMessage
--- A_DEL :: AMessage
+parseMessage :: Message -> Either ErrorType AMessage
+parseMessage msg = case B.words msg of
+  -- ["HELLO", key, ackMode] -> Right $ HELLO key $ ackMode am
+  ["REPLY", qInfo] -> REPLY <$> parseSmpQueueInfo qInfo
+  ["A_MSG", msgBody] -> Right $ A_MSG msgBody
```
epoberezkin, line 104: That probably needs to change using the same approach as in https://github.com/simplex-chat/simplex-messaging/blob/master/src/Simplex/Messaging/Agent/Transmission.hs#L303

But not sure about it, I didn't think about variable size bodies... Let's think if there is a better approach to storing messages in the database other than just ByteString... Maybe we can somehow store fields. Not sure. It should not hold merging.

src/Simplex/Messaging/Agent/Transmission.hs
```diff
@@ -91,13 +90,50 @@ data ACommand (p :: AParty) where
...
 data AMessage where
   HELLO :: VerificationKey -> AckMode -> AMessage
   REPLY :: SMPQueueInfo -> AMessage
   A_MSG :: MsgBody -> AMessage
 
--- A_ACK :: AgentMsgId -> AckStatus -> AMessage
--- A_DEL :: AMessage
+parseMessage :: Message -> Either ErrorType AMessage
+parseMessage msg = case B.words msg of
+  -- ["HELLO", key, ackMode] -> Right $ HELLO key $ ackMode am
```
epoberezkin, line 102: This should be easily fixable with `HELLO key <$> ackMode am` (ackMode am returns Either and `HELLO key` is a function that takes one parameter, so it can be used with fmap).

src/Simplex/Messaging/Agent/Transmission.hs
epoberezkin: This question depends on whether we want to use the same parsing/serialization here as in the protocol (I thought we would use the same) where the size needs to be used in the syntax so that the end of the message can be known with binary that follows and can be padded.

epoberezkin (dismissed):
It's great. There is only one minor comment about how to fix HELLO, but it can be addressed in the next PR. Message size in the parser (as in MSG/SEND parser - see the link) - def better in a separate PR.

epoberezkin (approved):

PR #18, opened 2021-01-17
parse agent messages with attoparsec

epoberezkin:


src/Simplex/Messaging/Agent/Transmission.hs
```diff
@@ -103,60 +107,82 @@ data SMPMessage
...
   HELLO :: VerificationKey -> AckMode -> AMessage
   REPLY :: SMPQueueInfo -> AMessage
   A_MSG :: MsgBody -> AMessage
+  deriving (Show)
 
 parseSMPMessage :: ByteString -> Either ErrorType SMPMessage
-parseSMPMessage _ = Left INTERNAL
+parseSMPMessage =
+  first (const $ SYNTAX errBadMessage)
+    . A.parseOnly (smpMessageP <* A.endOfLine <* A.endOfInput)
```
efim-poberezkin, line 121: reuse `parse` from #184 here?

src/Simplex/Messaging/Agent/Transmission.hs
```diff
@@ -103,60 +107,82 @@ data SMPMessage
...
+
+agentMessageP :: Parser AMessage
+agentMessageP = hello <|> reply <|> a_msg
+  where
+    hello = "HELLO " *> (HELLO <$> base64P <*> ackMode)
+    reply = "REPLY " *> (REPLY <$> smpQueueInfoP)
+    a_msg = do
+      size <- "MSG " *> A.decimal :: Parser Int
+      A_MSG <$> (A.endOfLine *> A.take size <* A.endOfLine)
+    ackMode = " NO_ACK" $> AckMode Off <|> pure (AckMode On)
```
efim-poberezkin, line 162: apply same approach as in `parseCommandP` (#288) to avoid excessive backtracking? or is it unnecessary here?

src/Simplex/Messaging/Agent/Transmission.hs
```diff
@@ -103,60 +107,82 @@ data SMPMessage
...
+  SMPQueueInfo <$> ("smp::" *> smpServerP <* "::") <*> (base64P <* "::") <*> base64P
+
+smpServerP :: Parser SMPServer
+smpServerP = SMPServer <$> server <*> port <*> msgHash
   where
-    messageHeader agentMsgId agentTimestamp previousMsgHash =
-      B.unwords [B.pack $ show agentMsgId, B.pack (formatISO8601Millis agentTimestamp), encode previousMsgHash]
+    server = B.unpack <$> A.takeTill (A.inClass ":# ")
+    port = A.char ':' *> (Just . show <$> (A.decimal :: Parser Int)) <|> pure Nothing
+    msgHash = A.char '#' *> (Just <$> base64P) <|> pure Nothing
```
efim-poberezkin, line 173: what is the purpose of `msgHash` in smp server? is it for chain verification purposes?

src/Simplex/Messaging/Agent/Transmission.hs
efim-poberezkin: also I guess same could be done for `smpMessageP` (#125). Again not sure if it's needed there, there're only two alternatives rn

efim-poberezkin (approved):
looks great to me, just a couple comments and questions

src/Simplex/Messaging/Agent/Transmission.hs
epoberezkin: no, it's a pre-shared hash of the key (or some other secret) for transport encryption between server and the client - there should be some ad-hoc or noise-based protocol to negotiate a symmetric key for the TCP session.

src/Simplex/Messaging/Agent/Transmission.hs
epoberezkin: good spot - I missed - will change in the next PR.

src/Simplex/Messaging/Agent/Transmission.hs
epoberezkin: I am wondering if it has any effect on parsing effectiveness/backtracking - not sure... The main reason for parseCommandP change was code readability (although that applies here as well) - I was actually thinking to just remove simple functions and push it to the main function...

epoberezkin: when we refactor in the end we need to bring these parts closer - they are too far - difficult to manage

PR #19, opened 2021-01-17
agent store: add update queue methods

efim-poberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -272,12 +272,12 @@ sendConfirmation ::
   SendQueue ->
   SMP.SenderKey ->
   m ()
-sendConfirmation c SendQueue {server, sndId} senderKey = do
+sendConfirmation c sndQ@SendQueue {server, sndId} senderKey = do
```
epoberezkin, line 275: rcvQ and sndQ are used I think - doesn't he warn about shadowing?

epoberezkin: can be just `rq` or `sq`

efim-poberezkin: I don't get warnings.. Where are they used?

efim-poberezkin: w/e, changed anyway

epoberezkin:
that's the bug I fixed in my PR - I've just re-run the tests

epoberezkin (approved):

PR #21, opened 2021-01-18
Sending messages end to end

epoberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -126,19 +128,45 @@ processCommand c@AgentClient {sndQ} (corrId, connAlias, cmd) =
...
+        sendMsg sq = do
+          sendAgentMessage sq $ A_MSG msgBody
+          -- TODO respond $ SENT aMsgId
+          respond OK
+
+    ackMessage :: AgentMsgId -> m ()
+    ackMessage _aMsgId =
+      withStore (`getConn` connAlias) >>= \case
+        SomeConn _ (DuplexConnection _ rq _) -> ackMsg rq
+        SomeConn _ (ReceiveConnection _ rq) -> ackMsg rq
```
epoberezkin, line 148: @efim-poberezkin just looking at this code (and above) I am wondering whether we even need this SomeConn type  (and also Connection type with DuplexConnection / ReceiveConnection / SendConnection constructors) or would be better off with the methods `getRcvQueue connAlias` and `getSndQueue connAlias` that would error with SEBadConnection if there is no required queue - it would have avoided these cases.

epoberezkin: Same as we ended up doing with updateRcvQueueStatus

efim-poberezkin: I agree, sounds reasonable. Another point is that some other methods seem to operate only with `connAlias`, with Store checking connection type later on. Might just simplify implementation for those Store methods as well

epoberezkin: which methods?

efim-poberezkin: for example `sendReplyQInfo` calling `addRcvQueue`, with connection type being checked only on the latter

efim-poberezkin (approved):

PR #24, opened 2021-01-20
Logging

epoberezkin:


epoberezkin:
Good:
- doesn't log when running tests
Bad
- logs out of order
- client got coupled to agent
- not clear on behalf of which client the command was sent (client doesn't have it)

Possible options
- maybe we should just use putStrLn or something like this
- moving logging to agent - it would spread around but it would also be clearer what triggered SMP command - so maybe it's better
- passing clientId to SMPClient - more coupling
- possibly better to split the AgentClient into a separate module to provide API to SMP commands - it already exists and agent  uses it internally - logging would be there then

efim-poberezkin (approved):

PR #25, opened 2021-01-21
code-organization

efim-poberezkin:


src/Simplex/Messaging/Agent/Transmission.hs
```diff
@@ -366,7 +357,7 @@ tGetRaw h = do
 
 tPut :: MonadIO m => Handle -> ATransmission p -> m ()
 tPut h (corrId, connAlias, command) =
```
epoberezkin, line 359: You could put `CorrId corrId` here, then you won't need `bs` below.

epoberezkin (approved):

PR #28, opened 2021-01-23
suspend and delete connection

epoberezkin:
this should be reviewed after "subscription"

### DIFF
--- a/src/Simplex/Messaging/Agent.hs
+++ b/src/Simplex/Messaging/Agent.hs
@@ -107,6 +107,8 @@ processCommand c@AgentClient {sndQ} (corrId, connAlias, cmd) =
     SUB -> subscribeConnection
     SEND msgBody -> sendMessage msgBody
     ACK aMsgId -> ackMessage aMsgId
+    OFF -> suspendConnection
+    DEL -> deleteConnection
   where
     createNewConnection :: SMPServer -> m ()
     createNewConnection server = do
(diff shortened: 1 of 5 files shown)


src/Simplex/Messaging/Agent.hs
```diff
@@ -163,6 +165,27 @@ processCommand c@AgentClient {sndQ} (corrId, connAlias, cmd) =
       where
         ackMsg rq = sendAck c rq >> respond OK
 
+    suspendConnection :: m ()
+    suspendConnection =
+      withStore (`getConn` connAlias) >>= \case
+        SomeConn _ (DuplexConnection _ rq _) -> suspend rq
+        SomeConn _ (ReceiveConnection _ rq) -> suspend rq
```
efim-poberezkin, line 172: `suspend rq` is so small I'm not sure it's worth having it as a separate function, might be more readable just duplicating a couple of calls

tests/AgentTests.hs
```diff
@@ -20,34 +23,79 @@ agentTests = do
...
-  ("3", "bob", Right OK) <- sendRecv alice ("3", "bob", "SEND :how are you?")
-  ("", "alice", Right (MSG _ _ _ _ "hello")) <- tGet SAgent bob
-  ("12", "alice", Right OK) <- sendRecv bob ("12", "alice", "ACK 0")
-  ("", "alice", Right (MSG _ _ _ _ "how are you?")) <- tGet SAgent bob
-  ("13", "alice", Right OK) <- sendRecv bob ("13", "alice", "ACK 0")
-  ("14", "alice", Right OK) <- sendRecv bob ("14", "alice", "SEND 9\nhello too")
-  ("", "bob", Right (MSG _ _ _ _ "hello too")) <- tGet SAgent alice
-  ("4", "bob", Right OK) <- sendRecv alice ("4", "bob", "ACK 0")
-  return ()
+(#:) :: Handle -> (ByteString, ByteString, ByteString) -> IO (ATransmissionOrError 'Agent)
```
efim-poberezkin, line 39: Maybe add small comments explaining what these functions do? It would draw focus from reading them to reading tests. I don't think it'd be that distasteful considering they're not common

efim-poberezkin (approved):

src/Simplex/Messaging/Agent.hs
epoberezkin: small functions is good - easier to add things in the future. I am starting to create them easier - it's easier than refactor large functions later

tests/AgentTests.hs
epoberezkin: will add in the next PR

PR #29, opened 2021-01-24
handle TCP disconnections (WIP)

epoberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -46,13 +47,17 @@ runSMPAgent cfg@AgentConfig {tcpPort} = do
       q <- asks $ tbqSize . config
       n <- asks clientCounter
       c <- atomically $ newAgentClient n q
-      logInfo $ "client " <> showText (clientId c) <> " connected to Agent"
+      logConnection c True
```
efim-poberezkin, line 50: Is it possible to create synonyms for data constructors? Doesn't seem to be from what I'm finding..
I.e. `Connected` = `True`, `Disconnected` = `False`

Or maybe create a type to make intension more explicit? E.g. `data ConnectionStatus = Connected | Disconnected`

efim-poberezkin: Nvm I guess, I'm probably overthinking it

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -67,32 +71,53 @@ newAgentClient cc qSize = do
...
     connectClient :: m SMPClient
     connectClient = do
       cfg <- asks $ smpCfg . config
-      liftIO (getSMPClient srv cfg msgQ)
+      liftIO (getSMPClient srv cfg msgQ clientDisconnected)
         `E.catch` \(_ :: IOException) -> throwError (BROKER smpErrTCPConnection)
 
+    clientDisconnected :: IO ()
+    clientDisconnected = do
+      removeServerSubs >>= mapM_ (mapM_ removeNotifySub)
```
efim-poberezkin, line 103: Is it okay that server subscriptions and subscribed connections are removed separately (i.e. not in a single `atomically`)? Maybe it'd be more proper to remove them in a single function returning `Maybe (Set ConnAlias)` and then have a separate function make a notification based on provided set?

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -152,12 +177,26 @@ subscribeQueue c rq@ReceiveQueue {server, rcvPrivateKey, rcvId} connAlias = do
...
+    addSub :: Maybe (Set ConnAlias) -> Set ConnAlias
+    addSub (Just cs) = S.insert connAlias cs
+    addSub _ = S.singleton connAlias
 
 removeSubscription :: AgentMonad m => AgentClient -> ConnAlias -> m ()
-removeSubscription c connAlias =
-  atomically . modifyTVar (subscribed c) $ M.delete connAlias
+removeSubscription AgentClient {subscrConns, subscrSrvrs} connAlias = atomically $ do
+  cs <- readTVar subscrConns
+  writeTVar subscrConns $ M.delete connAlias cs
```
efim-poberezkin, line 191: Is it possible to reuse some functionality between `clientDisconnected` and `removeSubscription`?

Also what is the reason that it is `writeTVar subscrConns $ M.delete connAlias cs` in `removeSubscription`, but `modifyTVar (subscrConns c) $ M.delete connAlias` in `removeSubscription`? Maybe add small comments explaining the intention?

src/Simplex/Messaging/Client.hs
```diff
@@ -78,35 +80,52 @@ data Request = Request
     responseVar :: TMVar (Either SMPClientError Cmd)
   }
 
-getSMPClient :: SMPServer -> SMPClientConfig -> TBQueue SMPServerTransmission -> IO SMPClient
+getSMPClient :: SMPServer -> SMPClientConfig -> TBQueue SMPServerTransmission -> IO () -> IO SMPClient
```
efim-poberezkin, line 83: I understand why `IO ()` is added, but it looks a bit hacky. Make a type synonym e.g. `type DisconnectHandler = IO ()`?

efim-poberezkin: On the other hand it's clear from arguments

epoberezkin: it's quite common to pass continuation like this - all bracket functions would accept `IO a` as a parameter, no need to create a synonym for it - `IO a` always means "something to do with side effects and that returns a as a result". There is no real difference between IO () as return type and IO () as parameter - both define an action that can be performed.

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: >Is it possible to reuse some functionality between clientDisconnected and removeSubscription?

maybe, not clear... any suggestion?

> Also what is the reason that it is writeTVar subscrConns $ M.delete connAlias cs in removeSubscription, but modifyTVar (subscrConns c) $ M.delete connAlias in removeSubscription? Maybe add small comments explaining the intention?

I use this cs in the first case to remove subscriptions from the servers, the second only modifies it.

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -152,12 +177,26 @@ subscribeQueue c rq@ReceiveQueue {server, rcvPrivateKey, rcvId} connAlias = do
...
-  atomically . modifyTVar (subscribed c) $ M.insert connAlias (server, rcvId)
+addSubscription c ReceiveQueue {server} connAlias = atomically $ do
+  modifyTVar (subscrConns c) $ M.insert connAlias server
+  modifyTVar (subscrSrvrs c) $ M.alter (Just . addSub) server
+  where
+    addSub :: Maybe (Set ConnAlias) -> Set ConnAlias
+    addSub (Just cs) = S.insert connAlias cs
+    addSub _ = S.singleton connAlias
 
 removeSubscription :: AgentMonad m => AgentClient -> ConnAlias -> m ()
```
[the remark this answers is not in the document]
epoberezkin: > If I'm reading this function correctly it doesn't remove server from subscrSrvrs if connection set becomes empty

it does: `if S.null cs' then Nothing else Just cs'`

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: although don't think it's a huge problem as it is only accessed by a single thread.

apps/client-test/Main.hs
```diff
@@ -0,0 +1,52 @@
...
+ping :: Handle -> IO ()
+ping h = do
+  putStrLn "connected to server"
+  forever $ do
+    putLn h "ping"
+    threadDelay 1000000
+    line <- getLn h
+    print line
+
+runTCPClient :: HostName -> ServiceName -> (Handle -> IO a) -> IO a
```
efim-poberezkin, line 26: Is there a way to make `runTCPClient` from Transport more generic? If not maybe move this one to Transport as well and rename the one there to `runTCPClientUnliftIO` to increase chances to avoid code divergence?

Also `startTCPClient` is very similar apart from throwing exception, but it doesn't have latest changes from Transport.

efim-poberezkin (approved):

PR #31, opened 2021-01-25
Refactor END notifications

epoberezkin:
… test END notification when server connection dies


epoberezkin:
Interesting... Initially I thought the issue was that I couldn't kill the server in the test was that `forkIO` is used instead of `forkIOWithUnmasked`, but later it turned out that it was because I was not killing client threads (which is only necessary when the server is run as a thread, not as a process). So I reverted... But, strangely, it's important in the CI - locally `forkIO` works but not in CI anymore...

efim-poberezkin (approved):

PR #34, opened 2021-01-27
agent sqlite: rework schema, make methods transactional, enable foreign keys

efim-poberezkin:


epoberezkin (approved):

efim-poberezkin:
`WITHOUT ROWID` enforces NOT NULL on every column of the PRIMARY KEY, but agreed to duplicate with explicit `NOT NULL` for clarity.
https://sqlite.org/withoutrowid.html#differences_from_ordinary_rowid_tables

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -7,78 +7,99 @@ module Simplex.Messaging.Agent.Store.SQLite.Schema where
...
-    CREATE TABLE IF NOT EXISTS servers
-      ( server_id INTEGER PRIMARY KEY,
-        host TEXT NOT NULL,
-        port TEXT,
-        key_hash BLOB,
-        UNIQUE (host, port)
-      )
+    CREATE TABLE IF NOT EXISTS servers(
+      host TEXT NOT NULL,
+      service_name TEXT NOT NULL,
```
epoberezkin, line 16: service_name can actually be NULL. We should not be defaulting it to 5223, in the future we will support hostnames with unspecified ports - they idea is to be able to fallback to say web-sockets if 5223 is not opened.

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -7,78 +7,99 @@ module Simplex.Messaging.Agent.Store.SQLite.Schema where
...
+      rcv_id BLOB NOT NULL,
+      conn_alias TEXT NOT NULL,
+      rcv_private_key BLOB NOT NULL,
+      snd_id BLOB,
+      snd_key BLOB,
+      decrypt_key BLOB NOT NULL,
+      verify_key BLOB,
+      status TEXT NOT NULL,
+      ack_mode INTEGER NOT NULL,
```
epoberezkin, line 36: should we remove it for now? don't think we use it yet

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
efim-poberezkin: even though SQLite wrongfully supports it (due to a bug in early versions), agreed to use empty strings / underscores and convert it to `Nothing` instead of NULLs in primary key

efim-poberezkin: another option would be to keep host & service_name in a single field, would require addition parsing though

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -29,109 +29,66 @@ import UnliftIO.STM
...
---   forM_ sndQId $ deleteSndQueue st
---   deleteConnection st connAlias
---   when (isNothing rcvQId && isNothing sndQId) $ throwError SEBadConn
+      deleteConnCascade dbConn connAlias
+
+  upgradeConnWithSndQueue :: SQLiteStore -> ConnAlias -> SendQueue -> m ()
+  upgradeConnWithSndQueue SQLiteStore {dbConn} connAlias sndQueue =
+    liftIO (updateRcvConnWithSndQueue dbConn connAlias sndQueue) >>= \case
+      Right () -> return ()
+      Left e -> throwError e
```
epoberezkin, line 85: I think it's liftEither

efim-poberezkin: Changed and pushed it, but after looking at it again wasn't happy about it. It's just immediately clear what the function returns w/t having to look at its type signature when using case - decided to revert the change

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
```diff
@@ -1,382 +1,403 @@
...
-  }
+-- ? replace with ToField? - it's easy to forget to use this
+serializePort_ :: Maybe ServiceName -> ServiceName
+serializePort_ = fromMaybe "_"
 
-type QueueRowId = Int64
+deserializePort_ :: ServiceName -> Maybe ServiceName
+deserializePort_ port
+  | port == "_" = Nothing
+  | otherwise = Just port
```
epoberezkin, line 50: In such cases it's better to use pattern matches than guards:

```
deserializePort_ "_" = Nothing
deserializePort_ port = Just port
```

or

```
deserializePort_ = \case
  "_" -> Nothing
  port -> Just port
```

guards are usually for conditions when pattern matches don't work

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
```diff
@@ -1,382 +1,403 @@
...
-      |]
-      srv
-    DB.queryNamed
-      conn
-      "SELECT server_id FROM servers WHERE host = :host AND port = :port"
-      [":host" := host, ":port" := port]
-  case r of
-    [Only serverId] -> return serverId
-    _ -> throwError SEInternal
+{- ORMOLU_DISABLE -}
```
epoberezkin, line 67: I can imagine what ormolu did :) I didn't know you can do it...

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
```diff
@@ -1,382 +1,403 @@
...
-  case r of
-    [smpServer] -> return smpServer
-    _ -> throwError SENotFound
+insertRcvQueue_ :: DB.Connection -> ReceiveQueue -> IO ()
+insertRcvQueue_ dbConn ReceiveQueue {..} = do
+  let port_ = serializePort_ $ port server
+  DB.executeNamed
+    dbConn
+    insertRcvQueueQuery_
+    [":host" := host server, ":port" := port_, ":rcv_id" := rcvId, ":conn_alias" := connAlias, ":rcv_private_key" := rcvPrivateKey, ":snd_id" := sndId, ":snd_key" := sndKey, ":decrypt_key" := decryptKey, ":verify_key" := verifyKey, ":status" := status]
```
epoberezkin, line 116: maybe multiline?

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
```diff
@@ -1,382 +1,403 @@
...
+    INSERT INTO connections
+      ( conn_alias, rcv_host, rcv_port, rcv_id, snd_host, snd_port, snd_id)
+    VALUES
+      (:conn_alias,     NULL,     NULL,   NULL,:snd_host,:snd_port,:snd_id);
+  |]
 
-getRcvQueueByRecipientId :: (MonadUnliftIO m, MonadError StoreError m) => SQLiteStore -> RecipientId -> HostName -> Maybe ServiceName -> m ReceiveQueue
-getRcvQueueByRecipientId st@SQLiteStore {conn} rcvId host port = do
+retrieveConnQueues :: DB.Connection -> ConnAlias -> IO (Maybe ReceiveQueue, Maybe SendQueue)
+retrieveConnQueues dbConn connAlias =
```
epoberezkin, line 179: maybe make this one without transaction and use withTransaction where needed?

epoberezkin (approved):

src/Simplex/Messaging/Agent.hs
```diff
@@ -290,11 +290,10 @@ newSendQueue (SMPQueueInfo smpServer senderId encryptKey) = do
         SendQueue
           { server = smpServer,
             sndId = senderId,
+            connAlias,
             sndPrivateKey,
             encryptKey,
             signKey,
-            -- verifyKey,
-            status = New,
-            ackMode = AckMode On
```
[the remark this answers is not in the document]
efim-poberezkin: yes, just to not be distracted by commented code

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
efim-poberezkin: I thought we'd better set up max_line_length for ormolu and hlint, apparently for ormolu though you're supposed to program the rule yourself

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
efim-poberezkin: `retrieveConnQueues` is exported, the point in adding `retrieveConnQueuesTransactionless_` was to keep the former transactional, and to have the latter reused inside other utility functions. The only point where it could be made transactional elsewhere is on SQLite.hs above but then it kind of ruins splitting logic and physical implementation between it and Util.hs

tests/AgentTests/SQLiteTests.hs
```diff
@@ -40,31 +41,46 @@ action `returnsResult` r = runExceptT action `shouldReturn` Right r
...
+  it "should throw error if foreign keys are enabled" $ \store -> do
+    let inconsistent_query =
+          [sql|
+            INSERT INTO connections
+              (conn_alias, rcv_host, rcv_port, rcv_id, snd_host, snd_port, snd_id)
+            VALUES
+              ("conn1", "smp.simplex.im", "5223", "1234", "smp.simplex.im", "5223", "2345");
+          |]
+    DB.execute_ (dbConn store) inconsistent_query
+      `shouldThrow` (const True :: Selector DB.SQLError)
```
efim-poberezkin, line 83: It'd be great if we could check the exact error code here instead of a general exception, but I didn't come up with implementation. Let's discuss if it's easy to add, maybe I'll just add TODO for future.
https://hackage.haskell.org/package/sqlite-simple-0.4.18.0/docs/Database-SQLite-Simple.html#t:SQLError
I believe it should be `ErrorConstraint` in this case

src/Simplex/Messaging/Agent.hs
```diff
@@ -15,31 +15,31 @@ module Simplex.Messaging.Agent
...
 import Simplex.Messaging.Server (randomBytes)
-import Simplex.Messaging.Transport
+import Simplex.Messaging.Transport (putLn, runTCPServer)
 import Simplex.Messaging.Types (CorrId (..), MsgBody, PrivateKey, SenderKey)
-import UnliftIO.Async
+import UnliftIO.Async (race_)
 import UnliftIO.Exception (SomeException)
 import qualified UnliftIO.Exception as E
-import UnliftIO.IO
+import UnliftIO.IO (Handle)
```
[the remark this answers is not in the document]
efim-poberezkin: will fix

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
efim-poberezkin: Fixed long lines manually for now

src/Simplex/Messaging/Agent/Store/SQLite/Util.hs
efim-poberezkin: renamed to `retrieveConnQueues_`

PR #37, opened 2021-02-13
Sign / verify SMP commands

epoberezkin:


src/Simplex/Messaging/Crypto.hs
```diff
@@ -0,0 +1,146 @@
...
+import Data.ByteString.Char8 (ByteString)
+import Database.SQLite.Simple as DB
+import Database.SQLite.Simple.FromField
+import Database.SQLite.Simple.Internal (Field (..))
+import Database.SQLite.Simple.Ok (Ok (Ok))
+import Database.SQLite.Simple.ToField (ToField (..))
+import Simplex.Messaging.Parsers (base64P)
+import Simplex.Messaging.Util (bshow, (<$$>))
+
+newtype PublicKey = PublicKey C.PublicKey deriving (Eq, Show)
```
efim-poberezkin, line 38: `C` refers to this module in other places, is it possible to `import qualified Crypto.PubKey.RSA as RSA`?

apps/dog-food/Main.hs
```diff
@@ -36,8 +36,9 @@ cfg :: AgentConfig
 cfg =
   AgentConfig
     { tcpPort = undefined, -- TODO maybe take it out of config
-      tbqSize = 16,
+      rsaKeySize = 256,
```
efim-poberezkin, line 39: maybe specify bytes as ```(2048 `div` 8)```? This would probably be more readable as crypto primitives are usually sized in bits

efim-poberezkin:
update cryptonite dependency and pin it in package.yml?

src/Simplex/Messaging/Crypto.hs
epoberezkin: renamed to R

apps/dog-food/Main.hs
epoberezkin: yes - good idea - updated everywhere, I was also confused initially

efim-poberezkin (approved):

PR #44, opened 2021-02-15
terminal UI

epoberezkin:


apps/dog-food/Main.hs
```diff
@@ -219,49 +218,4 @@ receiveFromAgent t c = forever . atomically $ do
       Disconnected _ -> set Nothing
```
efim-poberezkin, line 218: not in scope of this pr, but I still don't think it will be great to change active contact on each received message once you have more than one active conversation. Maybe we could have a separate command for that - e.g. "switchActiveContactToLatestSender" (wording could differ :), that would've been triggered by some specific key, say `Tab`. Let's consider it in future scope maybe, at least discuss it.. I just can see it being very annoying

apps/dog-food/ChatTerminal.hs
```diff
@@ -0,0 +1,321 @@
...
+import Simplex.Messaging.Transport (getLn, putLn)
+import qualified System.Console.ANSI as C
+import System.IO
+import Types
+
+data ChatTerminal = ChatTerminal
+  { inputQ :: TBQueue ByteString,
+    outputQ :: TBQueue ByteString,
+    activeContact :: TVar (Maybe Contact),
+    username :: TVar (Maybe Contact),
```
efim-poberezkin, line 35: might not be required currently, but why not have a separate `User` type? Wouldn't there be possible differences between them? Even now when reading Main line 122 `let user = Contact <$> name` I initially wondered why is user a contact.. maybe `Contact` can be part of a `User` - I understand user also needs his contact for sharing in future

epoberezkin: I am still treating it as a throw-away thing...

apps/dog-food/Main.hs
epoberezkin: let's park this conversation till will have this problem - then it will be easier to understand what is better.

apps/dog-food/ChatTerminal.hs
```diff
@@ -0,0 +1,329 @@
...
+      "\DEL" -> KeyBack
+      "\t" -> KeyTab
+      '\ESC' : _ -> KeyUnsupported
+      cs -> KeyChars cs
+
+    keyChars cs = do
+      c <- getChar
+      more <- hReady stdin
+      -- for debugging - uncomment this, comment line after: 
+      -- (if more then keyChars else \c' -> print (reverse c') >> return c') (c : cs)
```
efim-poberezkin, line 283: if you don't mind I'll keep this as a comment - I myself criticize commented code, but this will just get squashed and we'll have to recollect it, and it's just very convenient for testing. I think it's okay to keep this one line commented together with the comment explaining it's for debugging

efim-poberezkin (approved):
Looks great!

PR #45, opened 2021-02-16
agent: save messages

efim-poberezkin:
(acks not in scope of this pr)

**sending**
get id -> save to db -> send -> update status that sent -> receive notification -> update status that delivered

**receiving**
receive -> validate id and hash -> get id -> save to db -> acknowledge delivery to the server -> deliver to the client -> update status as sent to client -> send delivery confirmation to another agent via another channel -> update status as confirmed

**snd messages statuses**
created, sent, delivered

**rcv messages statuses**
received, acknowledged to broker, acknowledged to sender


src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -77,27 +77,59 @@ connections =
...
+      snd_msg_id INTEGER,
       message BLOB NOT NULL,
-      direction TEXT NOT NULL,
       msg_status TEXT NOT NULL,
-      PRIMARY KEY(agent_msg_id, conn_alias),
-      FOREIGN KEY(conn_alias) REFERENCES connections(conn_alias)
-    ) WITHOUT ROWID;
+      FOREIGN KEY (conn_alias) REFERENCES connections (conn_alias),
+      FOREIGN KEY (rcv_msg_id) REFERENCES rcv_messages (rcv_msg_id),
+      FOREIGN KEY (snd_msg_id) REFERENCES snd_messages (snd_msg_id),
```
epoberezkin, line 99: wouldn't it be better to have these IDs in queues? Particularly for non-duplex connections - they seem to belong there...

epoberezkin: ah - sorry - got confused... this is still separate tables design...

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -109,8 +107,13 @@ class Monad m => MonadAgentStore s m where
   removeSndAuth :: s -> ConnAlias -> m ()
   setRcvQueueStatus :: s -> ReceiveQueue -> QueueStatus -> m ()
   setSndQueueStatus :: s -> SendQueue -> QueueStatus -> m ()
+  -- ? make data kind out of AMessage so that we can limit parameter to AMessage A_MSG?
+  -- ? or just throw error / silently ignore other AMessage types?
```
epoberezkin, line 111: It actually seems like AMessage is the wrong type because it does not include the fields present in MSG constructor of ACommand - all the timestamps and IDs...

AMessage is something that we parse out of the SMP.MSG + additional field, but it may be that we just need a separate type that we save to the database that could be used to create agent MSG.

It seems like MessageDelivery is the type that should be saved to the database - it missed some fields but it is closest to what we need.

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -57,10 +57,10 @@ sndQueues =
       encrypt_key BLOB NOT NULL,
       sign_key BLOB NOT NULL,
       status TEXT NOT NULL,
-      PRIMARY KEY(host, port, snd_id),
-      FOREIGN KEY(host, port) REFERENCES servers(host, port),
-      FOREIGN KEY(conn_alias)
-        REFERENCES connections(conn_alias)
+      PRIMARY KEY (host, port, snd_id),
```
epoberezkin, line 60: we should start making style changes as separate PRs (guilty myself of not doing it)

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -77,27 +77,39 @@ connections =
       snd_host TEXT,
       snd_port TEXT,
       snd_id BLOB,
-      PRIMARY KEY(conn_alias),
-      FOREIGN KEY(rcv_host, rcv_port, rcv_id) REFERENCES rcv_queues(host, port, rcv_id),
-      FOREIGN KEY(snd_host, snd_port, snd_id) REFERENCES snd_queues(host, port, snd_id)
+      last_rcv_msg_id INTEGER,
+      last_snd_msg_id INTEGER,
```
epoberezkin, line 81: particularly here it was difficult to see what's changed...

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -77,27 +77,39 @@ connections =
...
 messages =
   [sql|
     CREATE TABLE IF NOT EXISTS messages(
-      agent_msg_id INTEGER NOT NULL,
+      agent_msg_id INTEGER PRIMARY KEY,
+      agent_timestamp TEXT NOT NULL,
       conn_alias TEXT NOT NULL,
-      timestamp TEXT NOT NULL,
+      rcv_msg_id INTEGER,
+      snd_msg_id INTEGER,
```
epoberezkin, line 96: why do we need two IDs? would one ID and direction be better? I mean it made sense if we had two tables and one to unify them, but now we have one?

epoberezkin: not sure :)

efim-poberezkin: yeah.. makes sense, that would be less messy. I don't know why I didn't revert them

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -97,8 +97,55 @@ messages =
...
+-- * investigate if it's possible to select and update in one query in sqlite
+-- ? deferred fks are in assumption it's more convenient to delete from 'messages' side using agent_msg_id,
+-- ? otherwise we could've just have unique constraints on 'messages' table and no foreign keys here
+rcvMessages :: Query
+rcvMessages =
+  [sql|
+    CREATE TABLE IF NOT EXISTS rcv_messages(
+      agent_msg_id INTEGER NOT NULL,
+      conn_alias TEXT NOT NULL,
+      rcv_msg_id INTEGER NOT NULL,
```
epoberezkin, line 122: rcv_timestamp?

epoberezkin: can be null for sent messages

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -97,8 +97,55 @@ messages =
...
+
+-- messages agent sends - snd_msg_id is id of the message sent / to be sent,
+-- as in its number in order of sending
+sndMessages :: Query
+sndMessages =
+  [sql|
+    CREATE TABLE IF NOT EXISTS snd_messages(
+      agent_msg_id INTEGER NOT NULL,
+      conn_alias TEXT NOT NULL,
+      snd_msg_id INTEGER NOT NULL,
```
epoberezkin, line 143: ah - you made two tables...

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
efim-poberezkin: rcv_timestamp and snd_timestamp in snd_messages are agent_timestamp in the main one. I agree it's not very clear, do you think it'd be better to separate agent_timestamp into two tables?

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -69,35 +69,128 @@ sndQueues =
...
+        DEFERRABLE INITIALLY DEFERRED
+    ) WITHOUT ROWID;
+  |]
+
+-- table containing metadata of messages agent receives
+-- * external_snd_id
+--   id of the message at sender, i.e. 'external_snd_id' corresponds to 'internal_snd_id' from the sender's side.
+-- * rcv_status
+--   one of [semantically]: "received", "acknowledged to broker", "acknowledged to sender", changed in this order.
+-- * broker_ack_ts
```
epoberezkin, line 119: maybe ack_brocker_ts to be more different from broker_ts?

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -69,35 +69,128 @@ sndQueues =
...
+--   ts of acknowledgement to broker, corresponds to "acknowledged to broker" status, should be null until that;
+--   do not mix up with 'broker_ts' - ts created at broker after broker receives the message from sender.
+-- * sender_ack_ts
```
epoberezkin, line 122: and respectively ack_sender_ts

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -69,35 +69,128 @@ sndQueues =
...
+--   ts of acknowledgement to sender, corresponds to "acknowledged to sender" status, should be null until that;
+--   do not mix up with 'external_snd_ts' - ts created at sender before sending (which corresponds to 'internal_ts').
+--
+-- the order of inserting rcv messages - in transaction do:
+-- 1. look up 'last_internal_msg_id' and 'last_internal_rcv_msg_id' from 'connections' table;
+-- 2. increment internal ids and insert into 'messages' table;
+-- 3. insert into 'rcv_messages' table;
+-- 4. update internal ids in 'connections' table - application is responsible for consistency.
```
epoberezkin, line 130: "application is responsible for consistency" - a bit confusing, we have FKs

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -69,35 +69,128 @@ sndQueues =
...
+  [sql|
+    CREATE TABLE rcv_messages(
+      conn_alias TEXT NOT NULL,
+      internal_rcv_id INTEGER NOT NULL,
+      internal_id INTEGER NOT NULL,
+      external_snd_id INTEGER NOT NULL,
+      external_snd_ts TEXT NOT NULL,
+      broker_id INTEGER NOT NULL,
+      broker_ts TEXT NOT NULL,
+      rcv_status TEXT NOT NULL,
```
epoberezkin, line 142: is this field needed given that we have two timestamps? It can be set in the app from timestamp fields

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
```diff
@@ -69,35 +69,128 @@ sndQueues =
...
+-- 3. insert into 'snd_messages' table;
+-- 4. update internal ids in 'connections' table - application is responsible for consistency.
+sndMessages :: Query
+sndMessages =
+  [sql|
+    CREATE TABLE snd_messages(
+      conn_alias TEXT NOT NULL,
+      internal_snd_id INTEGER NOT NULL,
+      internal_id INTEGER NOT NULL,
+      snd_status TEXT NOT NULL,
```
epoberezkin, line 174: same question here: do we need this field if we have two ts

Just asking - probably easier to filter with the status field

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
efim-poberezkin: not with `last_internal_msg_is`'s in 'connections'

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
efim-poberezkin: good point

src/Simplex/Messaging/Agent/Store/SQLite/Schema.hs
efim-poberezkin: yep, probably we don't need it - but you make a good point it'll be more convenient for filtering

efim-poberezkin: I vote let's keep those

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -84,33 +122,100 @@ instance Eq SomeConn where
...
-    agentMsgId :: Int,
-    timestamp :: UTCTime,
-    message :: AMessage,
-    direction :: QueueDirection,
-    msgStatus :: DeliveryStatus
+    -- | Monotonically increasing id of a message per connection, internal to the agent.
+    -- Preserves ordering between both received and sent messages.
+    internalId :: InternalId,
+    internalTs :: InternalTs,
+    msgBody :: MsgBody
```
epoberezkin, line 203: no need to change in this PR, but it feels like this structure misses the direction tag to be "a message". It is kind of stupid to add it of course, given that we have this tag on the level above as constructor. Let's merge and see how it is used and change later. Names can the be revisited later too.

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -84,33 +122,100 @@ instance Eq SomeConn where
...
+  | AcknowledgedToSender
+  deriving (Eq, Show)
+
+type AckBrokerTs = UTCTime
+
+type AckSenderTs = UTCTime
+
+-- | A message sent by the agent to a recipient.
+data SndMsg = SndMsg
+  { baseData :: MsgBase,
```
epoberezkin, line 173: let's call it `msgBase` then

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -84,33 +122,100 @@ instance Eq SomeConn where
...
+    externalSndTs :: ExternalSndTs,
+    brokerId :: BrokerId,
+    brokerTs :: BrokerTs,
+    rcvStatus :: RcvStatus,
+    -- | Timestamp of acknowledgement to broker, corresponds to `AcknowledgedToBroker` status.
+    -- Do not mix up with `brokerTs` - timestamp created at broker after it receives the message from sender.
+    ackBrokerTs :: AckBrokerTs,
+    -- | Timestamp of acknowledgement to sender, corresponds to `AcknowledgedToSender` status.
+    -- Do not mix up with `externalSndTs` - timestamp created at sender before sending,
+    -- which in its turn corresponds to `internalTs`.
```
epoberezkin, line 146: ... in sending agent

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -1,31 +1,52 @@
...
+  getConn :: s -> ConnAlias -> m SomeConn
+  getRcvQueue :: s -> SMPServer -> SMP.RecipientId -> m ReceiveQueue
+  deleteConn :: s -> ConnAlias -> m ()
+  upgradeRcvConnToDuplex :: s -> ConnAlias -> SendQueue -> m ()
+  upgradeSndConnToDuplex :: s -> ConnAlias -> ReceiveQueue -> m ()
+  setRcvQueueStatus :: s -> ReceiveQueue -> QueueStatus -> m ()
+  setSndQueueStatus :: s -> SendQueue -> QueueStatus -> m ()
 
+  -- Msg management
+  createRcvMsg :: s -> ConnAlias -> MsgBody -> ExternalSndId -> ExternalSndTs -> BrokerId -> BrokerTs -> m ()
```
epoberezkin, line 43: this method should return something - most likely InternalId - because the agent should sent to the client message with internally assigned ID (currently it just slaps 0). It can be in the next PR.

Client then would use this ID to refer to this message in other commands

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -1,31 +1,52 @@
...
+  createSndMsg :: s -> ConnAlias -> MsgBody -> m ()
```
epoberezkin, line 44: Same here - there should be `SENT` acknowledgement sent to the client that contains internal ID

epoberezkin (approved):
Solid work. I will re-review SQL part more attentively tomorrow but ok to merge.

src/Simplex/Messaging/Agent.hs
```diff
@@ -249,6 +248,12 @@ processSMPTransmission c@AgentClient {sndQ} (srv, rId, cmd) = do
             A_MSG body -> do
               logServer "<--" c srv rId "MSG <MSG>"
               -- TODO check message status
+              withStore $ \st -> createRcvMsg st connAlias body senderMsgId senderTimestamp srvMsgId srvTs
+              -- TODO either:
+              -- TODO   - pass MSG{.., msg = generatedRcvMsg} to notify,
+              -- TODO   - retrieve internalTs from generatedRcvMsg,
+              -- TODO   - pass recipientTs to createRcvMsg (second cleanest),
+              -- TODO   - remove m_recipient from MSG (might be the cleanest?).
               recipientTs <- liftIO getCurrentTime
```
epoberezkin, line 257: this timestamp should either be added inside store method or here and passed to store method (probably better)

src/Simplex/Messaging/Agent.hs
```diff
@@ -249,6 +248,12 @@ processSMPTransmission c@AgentClient {sndQ} (srv, rId, cmd) = do
             A_MSG body -> do
               logServer "<--" c srv rId "MSG <MSG>"
               -- TODO check message status
+              withStore $ \st -> createRcvMsg st connAlias body senderMsgId senderTimestamp srvMsgId srvTs
+              -- TODO either:
+              -- TODO   - pass MSG{.., msg = generatedRcvMsg} to notify,
+              -- TODO   - retrieve internalTs from generatedRcvMsg,
+              -- TODO   - pass recipientTs to createRcvMsg (second cleanest),
+              -- TODO   - remove m_recipient from MSG (might be the cleanest?).
```
epoberezkin, line 256: Internal ID is needed to refer to the messages, but we can change the syntax later

src/Simplex/Messaging/Agent/Store.hs
efim-poberezkin: good catch, for some reason it slipped my mind. will change in separate PR

src/Simplex/Messaging/Agent.hs
efim-poberezkin: let's discuss which approach to take here - will change in separate PR

PR #58, opened 2021-02-26
agent store: accept internal ts and return internal id

efim-poberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -164,10 +164,11 @@ processCommand c@AgentClient {sndQ} (corrId, connAlias, cmd) =
         _ -> throwError PROHIBITED -- NOT_READY ?
       where
         sendMsg sq = do
-          withStore $ \st -> createSndMsg st connAlias msgBody
+          senderTs <- liftIO getCurrentTime
+          senderId <- withStore $ \st -> createSndMsg st connAlias msgBody senderTs
           sendAgentMessage c sq $ A_MSG msgBody
-          -- TODO respond $ SENT aMsgId
-          respond $ SENT 0
+          -- ? change AgentMsgId to Int?
```
epoberezkin, line 170: Int64? yes

src/Simplex/Messaging/Agent.hs
```diff
@@ -164,10 +164,11 @@ processCommand c@AgentClient {sndQ} (corrId, connAlias, cmd) =
         _ -> throwError PROHIBITED -- NOT_READY ?
       where
         sendMsg sq = do
-          withStore $ \st -> createSndMsg st connAlias msgBody
+          senderTs <- liftIO getCurrentTime
```
epoberezkin, line 167: this timestamp should be the same as timestamp used in `mkAgentMessage` - it should be passed there, so they are the same

epoberezkin: and it should also be the same as internalTs I think for sent messages

src/Simplex/Messaging/Agent.hs
```diff
@@ -250,17 +251,22 @@ processSMPTransmission c@AgentClient {sndQ} (srv, rId, cmd) = do
...
+              recipientId <- withStore $ \st ->
+                createRcvMsg
+                  st
+                  connAlias
+                  body
+                  recipientTs
+                  senderMsgId
+                  senderTimestamp
+                  srvMsgId
+                  srvTs
```
epoberezkin, line 264: a single long line can be easier to read

src/Simplex/Messaging/Agent.hs
```diff
@@ -250,17 +251,22 @@ processSMPTransmission c@AgentClient {sndQ} (srv, rId, cmd) = do
...
               notify connAlias $
                 MSG
                   { m_status = MsgOk,
-                    m_recipient = (0, recipientTs),
+                    -- ? change AgentMsgId to Int?
```
epoberezkin, line 268: Int64

src/Simplex/Messaging/Agent.hs
efim-poberezkin: agreed.. it just goes beyond 120 length which becomes a bit too much.. for now w/e I guess, we have some signatures in Store that also exceed 120. Will return to it once we set up line length rules

epoberezkin (approved):

PR #59, opened 2021-02-27
SMP heartbeat to maintain the connection

epoberezkin:


src/Simplex/Messaging/Protocol.hs
```diff
@@ -71,11 +71,13 @@ data Command (a :: Party) where
   OFF :: Command Recipient
   DEL :: Command Recipient
   SEND :: MsgBody -> Command Sender
+  PING :: Command Sender
```
efim-poberezkin, line 74: not exclusive to sender? well I guess that's just previous SMP abstraction being stretched here.. whatever

but maybe add a separate `Agent` party?

efim-poberezkin: also could you separate `Broker` and "Agent" commands with comment headers please? ("Agent" commands are of `Recipient` & `Sender`, `Agent` if you add it as party. `Recipient` and `Sender` could be separate sections)

src/Simplex/Messaging/Crypto.hs
```diff
@@ -243,3 +245,13 @@ rsaPrivateKey pk =
       R.private_dQ = undefined,
       R.private_qinv = undefined
     }
+
```
efim-poberezkin, line 248: this file probably got into pr by a mistake

efim-poberezkin (approved):

src/Simplex/Messaging/Protocol.hs
epoberezkin: no problem if it is assigned to Sender. We do not create one TCP connection per queue, meaning that in one. TCP connection there can be both "Sender" and "Receiver" commands.

The point to make it belong to Sender is that if/when we add authentication, the server would reject Recipient commands without server user authentication.

src/Simplex/Messaging/Crypto.hs
epoberezkin: just didn't want to delete this function - let it be :)

src/Simplex/Messaging/Protocol.hs
```diff
@@ -64,14 +64,17 @@ type SenderId = QueueId
 type QueueId = Encoded
 
 data Command (a :: Party) where
+  -- Agent Recipient commands
```
epoberezkin, line 67: these are not agent commands

PR #60, opened 2021-02-27
Separate db connection for each TCP client connection

epoberezkin:


efim-poberezkin (approved):

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -41,9 +41,9 @@ import Data.Set (Set)
 import qualified Data.Set as S
 import Data.Text.Encoding
 import Data.Time.Clock
-import Numeric.Natural
 import Simplex.Messaging.Agent.Env.SQLite
 import Simplex.Messaging.Agent.Store
+import Simplex.Messaging.Agent.Store.SQLite
```
efim-poberezkin, line 46: redundant import

efim-poberezkin (approved):
Looks great, local tests on linux don't fail anymore either, it seems

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -45,10 +46,15 @@ data SQLiteStore = SQLiteStore
     dbConn :: DB.Connection
   }
 
-newSQLiteStore :: MonadUnliftIO m => String -> m SQLiteStore
-newSQLiteStore dbFilename = do
+createSQLiteStore :: MonadUnliftIO m => String -> m SQLiteStore
+createSQLiteStore dbFilename = do
+  store <- connectSQLiteStore dbFilename
```
efim-poberezkin, line 51: add the following TODO?

`TODO check that embedded SQLite library is compiled with threadsafe code [and if not raise an exception?]; this can be achieved by running "pragma COMPILE_OPTIONS;" and checking "SQLITE_THREADSAFE=1" or "SQLITE_THREADSAFE=2" is in the list of compile options (and there shouldn't be "SQLITE_THREADSAFE=0")`

this may be unnecessary if later we switch to TBQueue for queries across all threads, but at least would be a reminder until then

PR #63, opened 2021-02-28
test that sqlite library is compiled with threadsafe code

efim-poberezkin:


src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -55,7 +57,12 @@ createSQLiteStore dbFilename = do
 connectSQLiteStore :: MonadUnliftIO m => String -> m SQLiteStore
 connectSQLiteStore dbFilename = do
   dbConn <- liftIO $ DB.open dbFilename
-  return SQLiteStore {dbFilename, dbConn}
+  compileOptions <- liftIO (DB.query_ dbConn "pragma COMPILE_OPTIONS;" :: IO [[T.Text]])
+  let threadsafeOption = find (isPrefixOf "THREADSAFE=") (concat compileOptions)
+  case threadsafeOption of
+    Just "THREADSAFE=0" -> error "SQLite compiled with not threadsafe code" -- TODO warn instead
+    Just _ -> return SQLiteStore {dbFilename, dbConn} -- TODO info threadsafe
+    Nothing -> error "SQLite THREADSAFE compile option not found" -- should not happen
```
epoberezkin, line 65: Better put it into createSQLiteStore which is called once, connectSQLiteStore is called once per thread (twice for each TCP connection)

epoberezkin: If option not found - warn and continue
If option found and incorrect - we could exit or ask y/n to continue

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -55,7 +57,12 @@ createSQLiteStore dbFilename = do
 connectSQLiteStore :: MonadUnliftIO m => String -> m SQLiteStore
 connectSQLiteStore dbFilename = do
   dbConn <- liftIO $ DB.open dbFilename
-  return SQLiteStore {dbFilename, dbConn}
+  compileOptions <- liftIO (DB.query_ dbConn "pragma COMPILE_OPTIONS;" :: IO [[T.Text]])
+  let threadsafeOption = find (isPrefixOf "THREADSAFE=") (concat compileOptions)
```
epoberezkin, line 61: what does it actually return - a single item result that is array of texts? Just was wondering if it should be just maybeHead or something...

tests/AgentTests/SQLiteTests.hs
```diff
@@ -71,6 +73,12 @@ storeTests = withStore do
       describe "SndQueue exists" testCreateSndMsg
       describe "SndQueue doesn't exist" testCreateSndMsgNoQueue
 
+testCompiledThreadsafe :: SpecWith SQLiteStore
+testCompiledThreadsafe = do
+  it "should throw error if compiled sqlite library is not threadsafe" $ \store -> do
+    compileOptions <- DB.query_ (dbConn store) "pragma COMPILE_OPTIONS;" :: IO [[T.Text]]
+    shouldNotContain compileOptions [["THREADSAFE=0"]]
```
epoberezkin, line 80: better to write it as ```compileOptions `shouldNotContain` [["THREADSAFE=0"]]```

epoberezkin: but I don't think it is correct - try testing for what it definitely contains

src/Simplex/Messaging/Agent/Store/SQLite.hs
efim-poberezkin: array of arrays actually, that's why I do `concat`

tests/AgentTests/SQLiteTests.hs
efim-poberezkin: yep, right.. I was trying `compileOptions shouldNotContain [["THREADSAFE=0"]]` but missed backtics. But it is correct, I checked

src/Simplex/Messaging/Agent/Store/SQLite.hs
efim-poberezkin: connectSQLiteStore +

I remember why I put it in createSQLiteStore - I was trying to be safe - but after giving it a second thought I see my reasoning was incorrect.

nvm this, I'll fix it

epoberezkin (approved):

PR #64, opened 2021-02-28
streams rfc draft

epoberezkin:


rfcs/2021-02-28-streams.md
```diff
@@ -0,0 +1,47 @@
...
+
+For every fast/dense data transmission use a separate short-lived unidirectional SMP queue that the receiving agent would provision to receive such transmission and communicate queue address (server and queue ID).
+
+Such queues can be managed by agents via additional set of agent commands, without using connection abstraction, but instead using "stream" abstraction.
+
+The protocol to create and manage such stream would be the following:
+
+1. The sender client notifies the agent about the stream it has to create for a specific duplex connection (the connection must be duplex for it to be possible): `<conn_alias> SNEW <name> <size>` (where `conn_alias` is the existing duplex connection, `size` is the optional size for the data to be transferred; 0 for unbounded streams).
+2. The sender agent sends to the recipient agent the "envelope" `SNEW <name> <size>`
+3. The receiving agent would notify the client about availability of the stream by sending `<conn_alias> STREAM <name> <size>` notification.
```
efim-poberezkin, line 31: for consistency with later steps - should `name` be renamed to `stream_id`? or were these implied to be different entities?

epoberezkin: It is not well thought through. The idea was that the name is descriptive, but possibly it is not needed, as probably some additional information about the stream needs to be added by the clients as a normal message anyway, so that the client can decide to accept or not. In this case SNEW would not need to have the name at all, instead the agent would respond with stream ID. Same problem as with current connection name/ID.

I think we should postpone finalising it until chat protocol formalization.

epoberezkin: removed names

PR #66, opened 2021-03-03
agent store: explain use for various message Ids

efim-poberezkin:


epoberezkin (approved):

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -197,7 +201,9 @@ type DeliveredTs = UTCTime
 data MsgBase = MsgBase
   { connAlias :: ConnAlias,
     -- | Monotonically increasing id of a message per connection, internal to the agent.
-    -- Preserves ordering between both received and sent messages.
+    -- Internal Id is needed to track the order of the conversation (which can be different for the sender / receiver)
+    -- and address messages in commands. External [sender] Id cannot be used for this purpose
+    -- due to a possibility of implementation errors in different agents.
```
epoberezkin, line 206: and also because it does not include sent messages anyway, and we need a single sequence of sent and received messages.

efim-poberezkin: > and also because it does not include sent messages anyway

^ confused about what you mean here

> and we need a single sequence of sent and received messages

^ I'll return initial wording - "Preserves ordering ..." - in addition to current text

PR #67, opened 2021-03-03
agent store: organize sender and broker fields into tuples

efim-poberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -253,13 +253,15 @@ processSMPTransmission c@AgentClient {sndQ} st (srv, rId, cmd) = do
...
+              let m_broker_ = (srvMsgId, srvTs)
+              recipientId <- withStore $ createRcvMsg st connAlias body recipientTs m_sender_ m_broker_
               notify connAlias $
                 MSG
                   { m_status = MsgOk,
                     m_recipient = (recipientId, recipientTs),
-                    m_sender = (senderMsgId, senderTimestamp),
-                    m_broker = (srvMsgId, srvTs),
+                    m_sender = m_sender_,
+                    m_broker = m_broker_,
```
epoberezkin, line 264: you could have used `m_sender` and `m_broker` and then in the record you would not need `=`

efim-poberezkin: 🤦

epoberezkin (approved):

PR #76, opened 2021-03-30
automate releases

efim-poberezkin:


.github/workflows/build.yml
```diff
@@ -1,17 +1,60 @@
...
+          - os: macos-latest
+            cache_path: ~/.stack
+            stack_args: ""
+            artifact_rel_path: /bin/dog-food
+            asset_name: dog-food-macos-x86-64
+          - os: windows-latest
+            cache_path: C:/sr
+            stack_args: ""
+            artifact_rel_path: /bin/dog-food.exe
+            asset_name: dog-food-windows-x86-64
```
epoberezkin, line 57: add .exe?

epoberezkin (approved):

epoberezkin:
Suggest removing “chore” from name

.github/workflows/build.yml
efim-poberezkin: we'll have to rename it in install script anyway - adding .exe for windows seems similar to chmod +x for unix - maybe keep it the way it is?

PR #79, opened 2021-04-05
Read server keys from files or create if absent

epoberezkin:


apps/smp-server/Main.hs
```diff
@@ -1,18 +1,71 @@
...
-  putStrLn $ "Listening on port " ++ tcpPort cfg
-  runSMPServer cfg
+  (k, pk) <- readCreateKeys
+  B.putStrLn $ "SMP transport key hash: " <> publicKeyHash k
+  putStrLn $ "Listening on port " <> tcpPort cfg
+  runSMPServer cfg {serverKeyPair = (k, pk)}
+
+readCreateKeys :: IO C.KeyPair
+readCreateKeys = do
+  dir <- getAppUserDataDirectory "simplex"
```
efim-poberezkin, line 41: put these into /etc/simplex similar to ssh?

https://www.ssh.com/ssh/host-key

efim-poberezkin (approved):

apps/smp-server/Main.hs
efim-poberezkin: or better /etc/opt/simplex according to this
https://www.pathname.com/fhs/pub/fhs-2.3.html#OPTADDONAPPLICATIONSOFTWAREPACKAGES

PR #83, opened 2021-04-08
refactor chat terminal

epoberezkin:
The only changes are to split files to general logic and system-specific implementation - cross-platform "editor" mode will be in the next PR


efim-poberezkin (approved):
Good rearrangement

PR #84, opened 2021-04-08
automate changelogs

efim-poberezkin:


epoberezkin (approved):

.github/workflows/build.yml
```diff
@@ -16,12 +16,23 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Create release
         uses: softprops/action-gh-release@v1
         with:
-          fail_on_unmatched_files: true
+          body: ${{ steps.build_changelog.outputs.changelog }}
           files: |
             LICENSE
+          fail_on_unmatched_files: true
```
epoberezkin, line 35: why is this needed?

efim-poberezkin: Theoretically, to fail if there's no LICENSE (or other files we might add to the list above it). Practically, for some reason the action logs:
`Warning: Unexpected input(s) 'fail_on_unmatched_files', valid inputs are ['body', 'body_path', 'name', 'tag_name', 'draft', 'prerelease', 'files']`
even though it's clearly in their docs (and IIRC I've seen it in their code):
https://github.com/marketplace/actions/gh-release#-customizing

I've kept it so that if they fix it it would be guarding - it doesn't do any harm

PR #98, opened 2021-04-12
standard encoding for keys

epoberezkin:


apps/smp-server/Main.hs
```diff
@@ -1,13 +1,15 @@
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
 
 module Main where
 
 import Control.Monad (when)
-import Data.Attoparsec.ByteString.Char8 (Parser)
```
efim-poberezkin, line 6: Could we gradually switch to importing specific identifiers / module as qualified, please? (in this case the former but I'm just saying as a general suggestion)
It's just so much nicer to support afterwards w/t having to rely on external tools to figure out what comes where from.

package.yaml
```diff
@@ -12,6 +12,9 @@ extra-source-files:
   - README.md
 
 dependencies:
+  - ansi-terminal == 0.10.*
```
efim-poberezkin, line 15: Is this added by a mistake?

efim-poberezkin (approved):

apps/smp-server/Main.hs
```diff
@@ -1,13 +1,15 @@
...
+import Crypto.Store.PKCS8
+import Crypto.Store.X509
```
efim-poberezkin, line 8: Could we gradually switch to importing specific identifiers / module as qualified, please? (in this case the former but I'm just saying as a general suggestion)
It's just so much nicer to support afterwards w/t having to rely on external tools to figure out what comes where from.

src/Simplex/Messaging/Crypto.hs
```diff
@@ -139,29 +165,23 @@ aesKeySize = 256 `div` 8
...
+          d = R.private_d pk
+      if d * d < n
+        then loop
+        else pure (PublicKey k, mkPrivateKey pk)
+
+rsaPrivateSize :: PrivateKey k => k -> Int
+rsaPrivateSize = R.public_size . R.private_pub . rsaPrivateKey
+
+validKeyPair :: FullKeyPair -> Bool
+validKeyPair (PublicKey k, FullPrivateKey pk) = k == R.private_pub pk
```
efim-poberezkin, line 184: Maybe just remove this and only store private key? It's FullPrivateKey stored now anyway. Or do we need pubkey saved separately for re-retrieval if it was lost by the server owner? I believe we discussed it but I don't remember the reason.

efim-poberezkin:
"Encode keys" - lowercase title for consistency? - this now gets into changelog list

package.yaml
epoberezkin: Colours in markdown

apps/smp-server/Main.hs
epoberezkin: ok in this case, not as a principle though.

A common case when it is very inconvenient if you are using a lot of the imports and it either requires very long imports or creates a lot of visual noise.

PR #112, opened 2021-04-28
Updated ChatHelpInfo

markaleksanderh:


epoberezkin:
looks good - please merge - I will update it a bit later to reduce the height...
Probably cyan color is better too?

epoberezkin (approved):

PR #114, opened 2021-04-30
save keys as binary to db, remove legacy encoding

epoberezkin:


src/Simplex/Messaging/Crypto.hs
```diff
@@ -112,21 +110,21 @@ instance PrivateKey FullPrivateKey where
...
 
 instance IsString PublicKey where
-  fromString = parseString decodePubKey
+  fromString = parseString (decode >=> decodePubKey)
 
 parseString :: (ByteString -> Either String a) -> (String -> a)
 parseString parse = either error id . parse . B.pack
 
-instance ToField SafePrivateKey where toField = toField . serializePrivKey
+instance ToField SafePrivateKey where toField = toField . encodePrivKey
```
efim-poberezkin, line 121: to differentiate from `encode` / `decode` rename add prefix `binary_` to `encodePrivKey` / `encodePubKey`? - similar to parsers

efim-poberezkin: also maybe `import Data.ByteString.Base64 (decode, encode)`? had to go to ide for those

efim-poberezkin (approved):

PR #127, opened 2021-05-04
build SMP server on tags

epoberezkin:


.github/workflows/build.yml
```diff
@@ -81,20 +64,20 @@ jobs:
...
         with:
-          path: ${{ matrix.cache_path }}
+          path: ~/.stack
           key: ${{ matrix.os }}-${{ hashFiles('stack.yaml') }}
 
       - name: Build & test
         id: build_test
         run: |
-          stack build ${{ matrix.stack_args }}
+          stack test
```
efim-poberezkin, line 73: I was doing `stack build --test` even when it wasn't parameterized.. let's keep that - it's more explicit that it also builds

efim-poberezkin (approved):

PR #139, opened 2021-05-13
websockets transport

epoberezkin:


src/Simplex/Messaging/Client.hs
```diff
@@ -92,7 +93,7 @@ data SMPClientConfig = SMPClientConfig
   { -- | size of TBQueue to use for server commands and responses
     qSize :: Natural,
     -- | default SMP server port if port is not specified in SMPServer
-    defaultPort :: ServiceName,
+    defaultTransport :: (ServiceName, ATransport),
```
efim-poberezkin, line 96: update haskell doc

src/Simplex/Messaging/Client.hs
```diff
@@ -126,88 +127,90 @@ type Response = Either SMPClientError Cmd
...
+      bSize <- tcpTimeout `timeout` atomically (takeTMVar thVar)
+      pure $ case bSize of
+        Just (Right blockSize) -> Right c {action, blockSize}
+        Just (Left e) -> Left e
+        Nothing -> Left SMPNetworkError
+
+    useTransport :: (ServiceName, ATransport)
+    useTransport = case port smpServer of
+      Nothing -> defaultTransport cfg
+      Just "80" -> ("80", transport @WS)
```
efim-poberezkin, line 170: I'm not sure about hardcoding port in `useTransport`.. why not have `Transport` as field of `SMPServer`?

efim-poberezkin (approved):

src/Simplex/Messaging/Client.hs
epoberezkin: Which one?

src/Simplex/Messaging/Client.hs
epoberezkin: Discussed, as port 80 can only be WS

PR #141, opened 2021-05-15
readme fix

npoberezkin:
simplex-messaging.md reformulation of key pairs handling part


epoberezkin (approved):

protocol/simplex-messaging.md
```diff
@@ -266,7 +266,7 @@ Simplex Messaging Protocol:
 
   - One unique "public" key is used by the servers to authenticate requests to send the messages into the queue, and another unique "public" key - to retrieve the messages from the queue. "Unique" here means that each "public" key is used only for one queue and is not used for any other context - effectively, this key is not public and does not represent any participant identity.
 
-  - Both "public" keys are provided to the server by the queue recipient when the queue is created.
+  - Both recipient and sender "public" keys are provided to the server by the queue recipient. "Public" key `RK` is provided when the queue is created, public key `SK` is proviced when the queue is secured.
```
efim-poberezkin, line 269: I think we overuse quotation marks for public and private keys. It's okay first time when it's being explained they're not really public, but then it only confuses - they're technically public and private for crypto purposes

PR #142, opened 2021-05-15
groups agent protocol

epoberezkin:


rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+  A ->> AA: gidA GADD idAB<br>(idAB - conn alias A has for B)
+  
+  note over AA: initiate internal connection gidAB for B in group
+
+  AA ->> BA: via idAB: G_INV gidABinv gMeta gInfo<br>(gMeta contains number of members - TBC what else<br>should not contain date of creation or who created it)
+  BA ->> B: idBA GINV invID gMeta gInfo
+  B ->> BA: gidB? GJOIN invID<br>(could be also GREJECT invID?)
+  BA ->> B: gidB OK
+
+  BA ->> AA: establish internal connection gidBA (using gidABinv) for A in group<br>(intermal means that it is not visible to the clients. Separate namespace?)
```
efim-poberezkin, line 23: need clarification on internal vs direct connections..

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+ 
+  AA -> A: gidA GCON idAB
+
+  note over M, B: below steps happen for each existing member M
+
+  AA ->> BA: via gidBA: GM_ADD lvlAM midABM|_<br>(midABM - how A identifies member M to B,<br>lvlAM - connection level, 0 - direct, 1 - via smbdy etc.)
+  
+  note over BA: initiate connection idBM for direct messages and<br> internal gidBM for group messages between B and M
```
efim-poberezkin, line 31: why does A write to B via `gidBA`, not `gidAB`? why is direct connection created implicitly w/t client's decision?

efim-poberezkin: I'm struggling to understand notations a bit - is "internal" connection not a new "physical duplex" connection, but another way to refer to existing "direct" connection? If so maybe rename "internal connection" -> "group connection", "internal" is very misleading here. Or my understanding is incorrect

efim-poberezkin: even if "internal" connection is only logical, still A should use her own `gidAB` to write to B, not `gidBA`, correct?

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+  
+  BA ->> AA: via gidBA: GM_INV midABM|BAM idBMinv/gidBMinv<br>(midBAM - how B identifies member M to A<br>idBMinv/gidBMinv - invitations to connect for member M)
```
efim-poberezkin, line 33: `midABM` -> `midBAM`

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+
+  AA ->> MA: via gidAM: GM_NEW lvlAB midAMB idBMinv/gidBMinv<br>(midAMB - how A identifies B to member M,<br>lvlAB - connection level, 0 - direct etc.)
```
efim-poberezkin, line 35: why levels are needed?

rfcs/groups/groups.mmd
efim-poberezkin: idBMinv/gidBMinv - swap B and M? this is invitation for connection from M to B - wouldn't it be better to keep these indices consistent in sender to receiver direction, or I'm missing something?

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+
+  MA ->> BA: establish connection with idBM -> idMB
+  MA ->> BA: establish internal connection with gidBM -> gidMB
```
efim-poberezkin, line 38: getting confused again by notation - does `->` mean second simplex connection is established in another direction? should Ms and Bs be swapped here as well?

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+
+  MA ->> AA: via gidMA: G_CON midAMB|MAB (or G_ADD_ERR)
```
efim-poberezkin, line 40: should it be MBA instead of MAB?

efim-poberezkin:
struggling with these triplet ids. Why not have a "global" participant id per group? Race conditions could probably be solved by vote. i.e.

- A and B want both to add new member
- A and B both try to reserve new "global" id for new member (say ids monotonically increase, they both know the latest id so they reserve by the exact id itself)
- A gets to majority of Ms faster than B, when B gets to them he gets "id in process of reservation" reply, A reserves id, B has to retry
- A reports to all Ms she successfully reserved id
- A still has to send similar messages when new member connected and Ms connect with him, but even reserved but not used id doesn't hurt since it monotonically increases anyway

whether B is malicious or not and also reports he reserved id successfully can probably be solved by all Ms sending each other id was reserved by whom at them, and then voting

If A and B have equal amount of reservations they retry with small random timeout.

Idk, just a though.. probably it's harder than I think, but it'd give global participant id per group and probably simpler interactions after id reservation

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+  MA ->> M: gidM GCON idMB
+
+  BA ->> AA: via gidBA: G_CON midABM|BAM (or G_ADD_ERR)
```
efim-poberezkin, line 43: same - why does A has to know how B identifies her to M? shouldn't it be BMA?

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+  BA ->> B: gidB GCON idBM
+
+  note over A, BA: once all members were sent to B
+  AA ->> BA: via gidAB: G_ALL
```
efim-poberezkin, line 47: G_ALL -> GM_ALL

rfcs/groups/groups.mmd
```diff
@@ -0,0 +1,102 @@
...
+  BA ->> B: gidB GJOINED
+
+  note over A, AA: once all members reported connection
+  AA ->> A: gidA GALL idAB
```
efim-poberezkin, line 51: no GALL in document

rfcs/groups/groups.mmd
epoberezkin: "internal" - another duplex connection, but not addressable by agent client, and only used by the agent internally.

> why is direct connection created implicitly w/t client's decision?

Discussed, needs to be clarified in the doc - in short, to have it available for when it is needed, as it can be long to create (because of async comms).

rfcs/groups/groups.mmd
epoberezkin: Discussed - we will use a single unique ID per member of the group, to allow member mentions in group messages.

rfcs/groups/groups.mmd
epoberezkin: Discussed - they will not be shared, but tracked internally by each user.

rfcs/groups/groups.mmd
epoberezkin: it means the same duplex connection, but it has another ID for another user.

rfcs/groups/groups.mmd
epoberezkin: will be same ID

rfcs/2021-05-23-introduction.md
```diff
@@ -0,0 +1,36 @@
...
+- message `cIdBA ICONF invId infoM` - notification to confirm introduction
+- command `cIdBM? ILET invId` - accept offer to be introduced (response is `cIdBM OK`, followed by `ICON`)
+- message `cIdBM ICON invId` - confirmation that connection is established to both introduced parties
+- message `cIdAB IDONE cIdAM` - confirmation that connection is established to the introducer
+
+### Agent envelopes
+
+- `I_NEW extIntroIdM  infoM` - new introduction offered by introducer
+- `I_INV extIntroIdB idBMInv infoB` - invitation to join connection from B to M sent via A
+- `I_NEW extIntroIdB idBMInv infoB` - new introduction forwarded by introducer
```
npoberezkin, line 27: extIntroIdM? that's how it is on the diagram
Is structure of commands described anywhere? It's not completely clear, at least for me
I_NEW X-id idBMInv infoB:
I_NEW - agents command
extIntroIdB - ? 
idBMInv - conn alias that B has for M (but with inv, not sure what that means)
infoB - some information (profile?) about B

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+2) The recipient of this verification message can have fewer messages received. This situation can be resolved in several ways:
+- try to retrieve all messages from the queue that is behind. It might work, but it may be that the sender is simply trying to send messages because the network is down.
+- wait until the next integrity check and only report integrity violation if during the next integrity check they still cannot reconcile the previous integrity check. This is probably an acceptable compromise.
+
+2. Each agent sends message receipts to sending agents - message receipts would contain a signature of the message hashes. In the agent uses the same verification key for each member of the group, as considered below, these receipts can be re-broadcasted to other agents (again, grouping them in one message to avoid `(n-1)*n` messages) as a proof that the messages were delivered.
+
+Comparing with the first approach, there are pros and cons:
+- pros:
+  - the sender would only send such integrity check messages when they have send a message to all parties, thus avoiding the situation when some messages might have been not yet sent (or failed to send and they are retrying).
```
npoberezkin, line 44: Can't see how this + follows from previous paragraphs.
1 case - agent sends integrity check regularly
2 case - agent sends integrity check when it receives message

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+- cons:
+  - integrity check message mush contain a signature per member, so it would be substantially larger.
```
npoberezkin, line 46: And in first case integrity check contains ids and digests per member. 
And from this line`broadcast: %s"CHECK" SP memId1 ":" msgId1 ":" hash1 ":" status1 SP memId2 ":" msgId2 ":" hash2 ":" status2` it seems that digest contains hashs.
Is hash-signtures > (ids + hashs), why?
What is "signature of hash"?

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+  - agents must use the same verification key for all members in the group, complicating the group and connections management.
```
npoberezkin, line 47: Why does it complicate managment?

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+  - in the same way as in the first scenario, some delivery notifications can be arbitrarily delayed.
+
+With either approach, each agent should probably track all the receipts of all messages from all agents. 
+
+Overall the first approach seems better. It shows who received which messages and the only case where the lack of integrity would be reported if the same message is different or some messages are skipped to some recipients (i.e. they broadcast integrity violation for some of the senders).
```
npoberezkin, line 52: What does it mean "same message is different"?

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+- message `G:gId SENT [C:cId] msgId` - notification that the message is sent and its internal ID
+- message `G:gId MSG C:cId msgdata` - received group message from cId, msgdata is the same set of parameters as in `MSG`
+- command `G:gId ACK msgId` - acknowledge message reception by the client
+- message `G:gId RCVD [C:cId] msgId status` - message delivery notification
+- command `G:gId LEAVE` - leave the group
+- message `G:gId LEFT [C:cId]` - connection cId left the group
+- command `G:gId REM C:cId` - remove group member (response is `G:gId OK`, followed by `REMD` notification)
+- message `G:gId REMD C:cId [C:cId]` - member removed (who, by whom - if it's not the current user)
+- message `G:gId OUT C:cId` - you are removed
+- message `G:gId EMPTY` - all members left the group and it is now empty
```
npoberezkin, line 72: For what do we need this message? Isn't it just message REMD for all members (which are known to the client).

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+- command `C:cId INTRO G:gId gInfo` - invite existing connection to a group
+- message `C:cId REQ G:invID gInfo` - invitation to join the group
+- command `G:gId? ACPT G:invId` - accept invitation (response is `G:gId OK`)
+- message `G:gId CON C:cId` - 2 connections created with some group member (both for group and direct messages)
+- message `G:gId MEM [C:cId]` - connection created with all group members for a given member or current client
+- command `G:gId SEND msg` - send message to group
+- message `G:gId SENT [C:cId] msgId` - notification that the message is sent and its internal ID
+- message `G:gId MSG C:cId msgdata` - received group message from cId, msgdata is the same set of parameters as in `MSG`
+- command `G:gId ACK msgId` - acknowledge message reception by the client
+- message `G:gId RCVD [C:cId] msgId status` - message delivery notification
```
npoberezkin, line 66: Will this notification be result of integrity check? At least I don't see envelope that support this kind of message

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+- `LEFT` - notification that member left the group
+- `OUT` - you are removed from the group
+- `REM C:mid` - remove member mid from the group
+- `REMD C:mid` - confirmation that member is removed
+- `DEL` - group is deleted
+- `DELD` - confirmation that group is deleted
+
+## Protocol costs
+
+- Adding a member:
```
npoberezkin, line 91: What are these magic numbers '1', '4', '6'

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+  - DEL - (n-1)
+  - DELD - (n-1)
+  - total - `2n-2` messages
+
+- Group integrity verification:
+  - TODO
+
+## Questions
+
+1. Message verification keys. Agents use separate server and encryption key for each connection, but there can be a value of having the same message verification key used for all members in the group. E.g., a member can validate to other group members that the message was delivered to all group members by sending signed message receipts they receive from the agent. This would mean that for each message `n*(n-1)` messages will be send, although signed receipts can be grouped to reduce this number. It can be done periodically, rather than on each message, as described [here](https://signal.org/blog/private-groups/).
```
npoberezkin, line 122: `member can validate to other group members that the message was delivered to all group members by sending signed message receipts they receive from the agent`
do you mean e.g.
1) m1 sends msg to m2, m3
2) m2 sends signed receipt back to m1
3) m3 sends signed receipt back to m1
4) m1 sends both signed receipts to m2, m3
It doesn't look like message integrity schemes described before, where either m1 regularly sends digests (without messages from m2,m3) or only 2) 3) happens without 4)

rfcs/2021-05-23-broadcast.md
```diff
@@ -0,0 +1,50 @@
+# SMP agent broadcast
+
+## Problem
+
+Support agent message broadcast to multiple connections.
+
+It is done in ad-hoc way as part of the previous [groups proposal](./2021-05-15-groups.md) - this proposal defines broadcast as a separate agent primitive to simplify group management.
```
efim-poberezkin, line 7: link seems to be broken

rfcs/2021-05-23-broadcast.md
```diff
@@ -0,0 +1,50 @@
...
+A minimal protocol of additional client commands to create, manage and use broadcasts.
+
+From the point of view of the recipient this will look like a normal message, as if the sending agent executed multiple send commands (in fact, broadcast can be implemented by agent sending itself multiple SEND commands)
+
+### Commands and messages
+
+- command `B:bId? NEW` - create broadcast (response is `B:bId OK`, or `ERR` if broadcast already exists)
+- command `B:bId ADD C:cId` - add existing connection to a broadcast (response is `B:bId OK` or `ERR`, e.g. if connection already added or does not exist)
+- command `B:bId SEND msg` - broadcast message (response is multiple `C:cId SENT msgId` or ERR, separately for each connection)
+- message `B:bId SENT [C:bId] msgId` - notification that the message is sent to a specific or all recipients
```
efim-poberezkin, line 22: so is response `C:cId SENT msgId` or `B:bId SENT [C:bId] msgId`? feels like these are the same and should be aligned, right? or is the second notification after all per connection responses?

rfcs/2021-05-23-broadcast.md
```diff
@@ -0,0 +1,50 @@
...
+- command `B:bId REM C:cId` - remove connection from broadcast (response is `B:bId OK` or `ERR`)
+- message `B:bId EMPTY` - all connections were removed from the broadcast
```
efim-poberezkin, line 24: are two messages sent on the removal of last connection in the broadcast - `OK` and `EMPTY`? do we have abstractions in place for such behavior? maybe we don't need `EMPTY` message at all? - client should be responsible for managing this himself as well - he can do `LS`

rfcs/2021-05-23-broadcast.md
```diff
@@ -0,0 +1,50 @@
...
+- command `B:bId DEL` - delete broadcast (response is `B:bId OK`)
+- command `B:bId LS` - list connections in broadcast, response is `MEM space_separated_connections`
```
efim-poberezkin, line 26: `MEM...` -> `B:bId MEM...`? or if `B:bId` is not needed then should it also be not need in other responses?

rfcs/2021-05-23-broadcast.md
```diff
@@ -0,0 +1,50 @@
...
+- message `B:bId MS space_separated_connections`
+
+## Questions
+
+1. Should broadcast IDs use the same namespace as connection IDs (and as group IDs)? Having the same namespace for all abstractions that the agent can operate on can be helpful, as it can also allow implementing some queries to determine which type a given ID has, but it also increases implementation complexity.
+
+2. Given that this abstraction would be used as internal abstraction for groups (same as connections internal to the group), it might be better to implement "agent users", each with its own connection namespace. In this case agent would use itself as one of the users.
```
efim-poberezkin, line 33: not clear on why it might be better to implement "agent users"

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
+# SMP agent groups
```
efim-poberezkin, line 1: there's only one groups rfc now - remove "2"?

rfcs/2021-05-23-introduction.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Solution
+
+A protocol with commands and message envelopes to exchange the information between parties to establish connection.
+
+### Commands and messages
+
+Below commands are for the scenario when A introduces B to M.
+
+- command `C:idAB INTRO C:idAM infoM` - introduce connection cIdB to connection cIdM (response is `C:idAB OK`)
```
efim-poberezkin, line 17: I'd suggest rephrasing since it's not introduction yet until B accepts. Also using Ids with different indexes in command and explanation (AB vs B) is confusing, just refer to parties - B, M?
`introduce connection cIdB to connection cIdM` -> `**proposal to** introduce B to M`

efim-poberezkin: I mean it doesn't really matter if these internal rfcs have ambiguities in formulations, but later they'll be added to protocol

rfcs/2021-05-23-introduction.md
```diff
@@ -0,0 +1,41 @@
...
+- command `C:idBM? ACPT C:invId` - accept offer to be introduced (response is `C:idBM OK`, followed by `C:idBM CON`)
+- message `C:idBM CON` - confirmation that connection is established to both introduced parties
+- message `C:idAB CON C:idAM` - confirmation that connection is established to the introducer
+
+### Agent envelopes
+
+- `INTRO C:extIntroIdM infoM` - new introduction offered by introducer
+- `INV C:extIntroIdB prv:invBM infoB` - invitation to join connection from B to M sent via A
+- `REQ C:extIntroIdB prv:invBM infoB` - new introduction forwarded by the introducer
+- `CON C:extIntroId` - confirmation that the connection is established sent by both introduced parties to the introducer
```
efim-poberezkin, line 28: just trying to understand - these `extIntroIds` are needed only as identifiers for an introduction from A to B and M so that B and M can later acknowledge `CON` by it - correct?

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
+- message `G:gId RCVD [C:cId] msgId status` - message delivery notification
+- command `G:gId LEAVE` - leave the group
+- message `G:gId LEFT [C:cId]` - connection cId left the group
+- command `G:gId REM C:cId` - remove group member (response is `G:gId OK`, followed by `REMD` notification)
+- message `G:gId REMD C:cId [C:cId]` - member removed (who, by whom - if it's not the current user)
+- message `G:gId OUT C:cId` - you are removed
+- message `G:gId EMPTY` - all members left the group and it is now empty
+- command `G:gId DEL` - delete the group (response is `G:gId OK`)
+- message `G:gId DELD [C:cId]` - group deleted (by whom, if it's not the current user)
+- command `G:bId LS` - list connections in broadcast, response is `MEM space_separated_connections`
```
efim-poberezkin, line 75: in broadcast -> in group

rfcs/2021-05-23-groups2.md
```diff
@@ -0,0 +1,134 @@
...
...XIgTVxuXG4gIEFBIC0-PiBNQTogdmlhIGdpZEFNOiBSRU0gQzptaWRCXG4gIG5vdGUgb3ZlciBNQTogcmVtb3ZlIGdpZE1CXG4gIE1BIC0-PiBBQTogdmlhIGdpZE1BOiBSRU1EIEM6bWlkQlxuICBNQSAtPj4gTTogRzpnaWRNIFJFTUQgQzppZE1CIEM6aWRNQTxicj4oQiByZW1vdmVkIGJ5IEEpXG5cbiAgbm90ZSBvdmVyIEEsIEFBOiBvbmNlIGFsbCBtZW1iZXJzIHJlbW92ZWQgQlxuXG4gIEFBIC0-PiBBOiBHOmdpZEEgUkVNRCBDOmlkQUI8YnI-KEIgcmVtb3ZlZCBieSB0aGlzIGFnZW50KVxuXG4gIG5vdGUgb3ZlciBNLCBCOiA0Yy4gQSBkZWxldGVzIGdyb3VwXG4gIEEgLT4-IEFBOiBHOmdpZEEgREVMXG4gIEFBIC0-PiBBOiBHOmdpZEEgT0tcbiAgXG4gIEFBIC0-PiBCQTogdmlhIGdpZEFCOiBERUxcbiAgbm90ZSBvdmVyIEJBOiByZW1vdmUgYWxsIGdyb3VwIGNvbm5lY3Rpb25zIGFuZCBtZXNzYWdlc1xuICBCQSAtPj4gQjogRzpnaWRCIERFTEQgQzppZEJBPGJyPihncm91cCBkZWxldGVkIGJ5IEEpXG4gIEJBIC0-PiBBQTogdmlhIGdpZEJBOiBERUxEXG4gIEFBIC0-PiBBOiBHOmdpZEEgREVMRCBDOmlkQUJcblxuICBBQSAtPj4gTUE6IHZpYSBnaWRBTTogREVMXG4gIG5vdGUgb3ZlciBNQTogcmVtb3ZlIGFsbCBncm91cCBjb25uZWN0aW9ucyBhbmQgbWVzc2FnZXNcbiAgTUEgLT4-IE06IEc6Z2lkTSBERUxEIEM6aWRNQTxicj4oZ3JvdXAgZGVsZXRlZCBieSBBKVxuICBNQSAtPj4gQUE6IHZpYSBnaWRNQTogREVMRFxuICBBQSAtPj4gQTogRzpnaWRBIERFTEQgQzppZEFNXG5cbiAgQUEgLT4-IEE6IEc6Z2lkQSBERUxEPGJyPihncm91cCBkZWxldGVkIGJ5IHRoaXMgYWdlbnQgLSBhbGwgY29uZmlybWVkKVxuIiwibWVybWFpZCI6e30sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)
```
efim-poberezkin, line 126: this link appears to be broken. also does it duplicate the first one?

rfcs/2021-05-23-broadcast.md
efim-poberezkin: nvm, I guess I got it when reading groups :)

rfcs/2021-05-23-groups2.md
epoberezkin: it's orthogonal question, I think it should be regular, irrespective of  whether agent sends what it receives or confirmations that other received (the question asked.

I think it should be merged without answering this question and refined later.

On regular integrity check, I think it might be a good idea to regularly send PING message between agents (say every hour or more/less frequently) to continuously maintain connection availability.

rfcs/2021-05-23-groups2.md
epoberezkin: hash === digest, these are synonyms.

"signature" is RSA-PSS cryptographic signature algorithm applied to digest.

rfcs/2021-05-23-groups2.md
epoberezkin: because it doesn't allow treating connections as black boxes that generate their own keys, it requires breaking the abstraction boundary and passing the same key from the outside of connection.

It effectively creates a long-term group-wide form of identity which is something that may need to be rotated in some scenarios and in general complicates things.

We already have a shared id per member, adding a shared key adds complexity.

rfcs/2021-05-23-groups2.md
epoberezkin: same message by order number in the sequence can have different content - can be clarified.

rfcs/2021-05-23-groups2.md
epoberezkin: possibly not needed as a notification, possibly can be used to indicate that it's impossible to send the message to the group, but it probably should be an error, possibly can be used to reply to `LS` command, although `MS` without members would do as well.

In current broadcast implementation I did not include EMPTY for these reasons, as it may be unnecessary.

rfcs/2021-05-23-groups2.md
epoberezkin: The envelope for notifications of message delivery is part of connection protocol, it is not needed separately for groups, as group communication simply uses dedicated connections.

rfcs/2021-05-23-groups2.md
epoberezkin: the number of messages transmitted to connect members, 4 is from introduction protocol, 6 is total.

rfcs/2021-05-23-groups2.md
epoberezkin: > It doesn't look like message integrity schemes described before, where either m1 regularly sends digests (without messages from m2,m3) or only 2) 3) happens without 4)

it is, roughly one of two schemes.

rfcs/2021-05-23-broadcast.md
epoberezkin: yes, it was renamed to a newer date, needs to be fixed

rfcs/2021-05-23-broadcast.md
epoberezkin: yes, didn't settle yet on what is better.

In the current implementation it is `C:cId SENT msgId`, but probably it is better to make it `B:bId SENT C:bId msgId` and then once sent to all `B:bId SENT msgId`. Although maybe not :)

There definitely should be notification per connection and then once all delivered per broadcast.

That, by the way, requires that the agent client, when it is coded, understands that one command can have multiple responses with the same correlation ID and can collect them all prior to returning the function result - SMP protocol does not have this problem - for command responses that do not depend on other agents and only depend on the servers.

rfcs/2021-05-23-broadcast.md
epoberezkin: Same comment as above. Maybe not needed. Can be helpful though to avoid the need to check group membership continuously...

rfcs/2021-05-23-broadcast.md
epoberezkin: yes, it's just left out.

rfcs/2021-05-23-broadcast.md
epoberezkin: > not clear on why it might be better to implement "agent users"

we need connections invisible to clients. Previously we had the idea that clients can have some form of authentication to an agent - "users" - in which case the agent itself can be such user. Just an idea.

rfcs/2021-05-23-introduction.md
epoberezkin: yes

rfcs/2021-05-23-groups2.md
epoberezkin: links are fixed now - they are to the same diagram (just one HTML, another svg)

PR #143, opened 2021-05-17
support permanent connection link ("contact") in SMP agent protocols

epoberezkin:


rfcs/2021-05-17-open-connection.md
```diff
@@ -0,0 +1,50 @@
...
+- an open invitation to connect to a person - e.g. QR code/invitation link on a person's website, or passed from one person to another.
+- part of the solution for public DNS-based addresses (when a directory server would map address in some domain name#example.tld to an open invitation).
+
+## Solution
+
+A minimal change to an SMP protocol - a dedicated unsecured SMP queue is used to receive invitations to connect: instead of the initial KEY message with the KEY to secure the queue, an additional INV message will be used to send an invitation to the queue of the sender, effectively using an unsecured SMP queue as an out-of-band channel for creating SMP queue.
+
+Additional commands in SMP agent protocol:
+
+`conn_alias? OPEN` (or `PUB`, `NEWPUB`, tbc) - to create an "open"/"public" queue, the response is an invitation in a different format (TBC):
```
efim-poberezkin, line 18: I like `NEWPUB` the best

rfcs/2021-05-17-open-connection.md
```diff
@@ -0,0 +1,50 @@
...
+  - should allow multiple servers (probably the original invitation should be extended to support it)
+  - should have a marker to indicate it's an open/public queue (probably the original invitation should be extended to include an invitation type).
+
+e.g. `smp:<queue_type>::<server1>/<queue_id1>,<server2>/<queue_id2>::<key1>`
```
efim-poberezkin, line 22: rename `<queue_type>` -> `<invitation_type>`? I understand it's for substitution anyway, but if we add it to protocol I think it's best renamed, otherwise it's ambiguous

rfcs/2021-05-17-open-connection.md
```diff
@@ -0,0 +1,50 @@
...
+
+BA ->> B: cidBA CON
+AA ->> A: cidAB CON
+```
+
+That protocol requires addressing the current problem when an invitation cannot be accepted when the party that generated the invitation is not online.
+
+Questions.
+
+1. Do we need to differentiate the semantics of the invitation on the syntax level, or should we allow to just manage it outside of protocol when the receiving agent decides which SMP messages to accept and which to ignore (KEY / INV).
```
efim-poberezkin, line 50: I think it's better to keep on syntax level, so as is, because although you called this invitation and it looks similar, it's very different in terms of semantics and is useless w/t continuing with regular invitations. So it's not like they're alternatives of the same, but rather just use similar concepts / terminology in very different contexts.

efim-poberezkin:
Overall looks very good to me, very clear

efim-poberezkin (approved):

PR #144, opened 2021-05-20
SMP server CLI commands

epoberezkin:


apps/smp-server/Main.hs
```diff
@@ -173,8 +173,8 @@ fileExists path = do
   exists <- liftIO $ doesFileExist path
   unless exists . throwE $ "file " <> path <> " not found"
 
-deleteFileIfExists :: FilePath -> IO ()
-deleteFileIfExists path = doesFileExist keyPath >>= (`when` removeFile path)
+deleteIfExists :: FilePath -> IO ()
+deleteIfExists path = doesFileExist keyPath >>= (`when` removeFile path)
```
npoberezkin, line 177: should we maybe write:
`doesFileExist path >>= ...`

epoberezkin: fixed

apps/smp-server/Main.hs
```diff
@@ -52,77 +52,129 @@ logDir = "/var/opt/simplex"
...
+          putStrLn "Server was initialized. Start it with `smp-server start` command"
+          printConfig cfg
+    ServerStart ->
+      runExceptT (getConfig opts) >>= \case
+        Right cfg -> runServer cfg
+        Left e -> do
+          putStrLn $ "Server is not initialized: " <> e
+          putStrLn "Initialize server with `smp-server init` command"
+          exitFailure
+    ServerDelete -> do
```
efim-poberezkin, line 80: can this be run at arbitrary moment, i.e. when server is started? if yes, will server just stop writing to store log after its deletion w/t crashing? if no, should there be a check that server is not started?

epoberezkin: I don’t hunk we should worry about it - it’s for the users who know why thaw doing.

npoberezkin (approved):

PR #149, opened 2021-05-25
turn off smp server websockets support in linode script

npoberezkin:


epoberezkin (approved):
I trust you it works :)

PR #150, opened 2021-05-26
add DigitalOcean SMP server Droplet configuration

npoberezkin:


scripts/smp-server-droplet/files/opt/simplex/on_first_login.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+printf "SMP server is nearly ready\n"
+printf "Do you want to persist SMP queues to append only log?\n(enables restoring queues info upon server restart)\nprint 1/2 for yes/no\n"
```
efim-poberezkin, line 4: why is it `1/2` not `y/n`?

scripts/smp-server-droplet/marketplace-image.json
```diff
@@ -0,0 +1,53 @@
+{
+  "variables": {
+    "token": "{{env `DIGITALOCEAN_TOKEN`}}",
+    "image_name": "smp-server-20-04-snapshot-{{timestamp}}",
```
efim-poberezkin, line 4: can we either have `ubuntu-20-04` in image name, or no `20-04` at all - it's kind of ambiguous to what these numbers refer if you don't say `ubuntu` - first thing I'd think it's version of smp-server. I think just `smp-server` w/t `ubuntu-20-04` might be better, it's not like we're giving alternatives on digital ocean

scripts/smp-server-droplet/scripts/03-prep.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# retrieve latest release info and download smp-server executable
+echo "downloading SMP server release $release_tag"
+curl -s https://api.github.com/repos/simplex-chat/simplexmq/releases/tags/{$release_tag} \
+| jq -r '.assets[].browser_download_url | select(test("smp-server-ubuntu-20_04-x86-64"))' \
+| tr -d \" \
+| wget -qi -
+
+echo "preparing for SMP server initiaization on first login"
+# move smp-server executable to /opt/simplex/bin
```
efim-poberezkin, line 11: don't think comment is needed here, commands below are self-explanatory

scripts/smp-server-droplet/scripts/03-prep.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# retrieve latest release info and download smp-server executable
+echo "downloading SMP server release $release_tag"
```
efim-poberezkin, line 4: comment and echo explain the same thing

scripts/smp-server-droplet/scripts/03-prep.sh
```diff
@@ -0,0 +1,46 @@
...
+| tr -d \" \
+| wget -qi -
+
+echo "preparing for SMP server initiaization on first login"
+# move smp-server executable to /opt/simplex/bin
+mkdir -p /opt/simplex/bin
+mv smp-server-ubuntu-20_04-x86-64 /opt/simplex/bin/smp-server
+chmod +x /opt/simplex/bin/smp-server
+
+# add /opt/simplex/bin to $PATH
```
efim-poberezkin, line 16: same here

efim-poberezkin:
rename folder `smp-server-droplet` -> `smp-server-digitalocean-droplet`?

efim-poberezkin:
what of these files are custom? e.g. `99-img-check.sh` and `90-cleanup.sh` (maybe some others as well) look very digital ocean generic - do we have to version them? does digital ocean integrate with github and it's convenient to have them all in one place? if there's no integration w/ github and they're not custom maybe it's better to remove them?

scripts/smp-server-droplet/files/opt/simplex/on_first_login.sh
npoberezkin: changed

scripts/smp-server-droplet/marketplace-image.json
npoberezkin: named `simplex-chat-server-snapshot`, better than abbreviations (this name will be seen by users if I'm not mistaken)

scripts/smp-server-droplet/scripts/03-prep.sh
npoberezkin: removed

scripts/smp-server-droplet/scripts/03-prep.sh
npoberezkin: changed

scripts/smp-server-droplet/scripts/03-prep.sh
npoberezkin: removed

npoberezkin:
> 
> 
> rename folder `smp-server-droplet` -> `smp-server-digitalocean-droplet`?



> 
> 
> rename folder `smp-server-droplet` -> `smp-server-digitalocean-droplet`?

renamed

npoberezkin:
> 
> 
> what of these files are custom? e.g. `99-img-check.sh` and `90-cleanup.sh` (maybe some others as well) look very digital ocean generic - do we have to version them? does digital ocean integrate with github and it's convenient to have them all in one place? if there's no integration w/ github and they're not custom maybe it's better to remove them?

- `what of these files are custom` they are generic
- `do we have to version them?` they're needed to automate process of droplet cleaning and verification before snapshot creation
- `does digital ocean integrate with github and it's convenient to have them all in one place` didn't understand. From fast googling digital ocean has some integration, but I haven't tried it. Anyway we'll need to store it on github, digital ocean only stores snapshots (not sure in which form) for droplet creation, not pure scripts.
- `maybe it's better to remove them` again we need them:)

README.md
```diff
@@ -77,9 +77,20 @@ To deploy SMP server on [Linode](https://www.linode.com/):
 
 Please submit an [issue](https://github.com/simplex-chat/simplexmq/issues) if any problems occur.
 
-## 🚧 Deploy SMP server on DigitalOcean 🚧
+[<img alt="digitalocean" src="./img/digitalocean.svg" align="right" width="200">](https://cloud.digitalocean.com)
```
efim-poberezkin, line 80: align linode and digitalocean image links both to lead either to main site or to specific deployment?

README.md
```diff
@@ -77,9 +77,20 @@ To deploy SMP server on [Linode](https://www.linode.com/):
...
 
-Coming soon.
+## Deploy SMP server on DigitalOcean
+
+To deploy SMP server on [DigitalOcean](https://cloud.digitalocean.com):
+- Create a DigitalOcean account or login with an already existing one;
```
efim-poberezkin, line 85: make punctuation consistent - let's use dots like in other lists

README.md
```diff
@@ -77,9 +77,20 @@ To deploy SMP server on [Linode](https://www.linode.com/):
...
+- Click [create Droplet](https://cloud.digitalocean.com/droplets/new?size=s-1vcpu-1gb&region=fra1);
+- Choose the region and plan according to your requirements (for regular use cheapest Regular option should be sufficient).
```
efim-poberezkin, line 87: `(for regular use Regular option...)` sounds like a tautology, just `(Regular should be sufficient)`

README.md
```diff
@@ -77,9 +77,20 @@ To deploy SMP server on [Linode](https://www.linode.com/):
...
+- Provide ssh key to be able to connect to your Droplet via ssh. This step is required, because you will need to login to your Droplet and get a SMP server public key hash (`hash` later) either from the welcome message or from the file `/etc/opt/simplex/simplex.conf` on your Droplet after SMP server starts.
```
efim-poberezkin, line 88: again, it's implied that providing an _ssh key_ gives you ability to connect via _ssh_ and not some other protocol. why not just `Provide ssh key to be able to connect to your Droplet.`?

README.md
```diff
@@ -77,9 +77,20 @@ To deploy SMP server on [Linode](https://www.linode.com/):
...
+- Create your Droplet.
```
efim-poberezkin, line 89: `Create your Droplet` -> `Confirm Droplet creation`? I'd move it to the previous step - you're focusing too much on following sites' UIs letter to letter - if DigitalOcean slightly adjust their forms (for example by adding one more choice) this tutorial is out of sync. E.g.

```Provide ssh key and confirm Droplet creation. Ssh key is required for you to login to the Droplet and retrieve and SMP server public key hash either from the welcome message or `/etc/opt/simplex/simplex.conf` after server start```

File: README.md (hunk 9 above)
Comment (npoberezkin) [reply README.md:88]: done

File: scripts/smp-server-digitalocean-droplet/scripts/03-init-start.sh [hunk 11]
```diff
@@ -0,0 +1,50 @@
...
+WantedBy=multi-user.target
+
+EOT
+chmod 644 /etc/systemd/system/smp-server.service
+
+echo "initializing SMP server"
+init_res="/etc/opt/simplex/simplex.conf"
+init_res_tmp="/etc/opt/simplex/simplex.tmp"
+smp-server init -l > $init_res
+tail -n +2 "$init_res" > "$init_res_tmp" && mv "$init_res_tmp" "$init_res"
```
efim-poberezkin, line 44: I think this is kind of hacky.. This only keeps last 2 lines of logs which happen to be public key, and then presents it as conf file, right? Usually conf files are kind of static and definitely are not pruned logs.. - basically what we have as smp-server.ini. Having seconf "conf" is confusing.. Also, what is the other line? Can we only extract pub key hash and store it in file "pub_hash" (so move grep from above here basically)?

scripts/smp-server-digitalocean-droplet/scripts/90-cleanup.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
efim-poberezkin, line 1: I still didn't get why we version these scripts, let's discuss offline

scripts/smp-server-digitalocean-droplet/scripts/03-init-start.sh
epoberezkin: we can grep the line with the hash, rather than assume it's a specific log line.

scripts/smp-server-digitalocean-droplet/files/opt/simplex/on_login.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+ip_address=$(hostname -I | awk '{print$1}')
+hash=$(cat /etc/opt/simplex/simplex.conf | grep hash: | cut -f2 -d":" | xargs)
```
epoberezkin, line 4: at some point I thought that it might be good to have hash as a parameter of .ini file so in addition to it all being in one location, the server would validate it.

@npoberezkin @efim-poberezkin

scripts/smp-server-digitalocean-droplet/scripts/90-cleanup.sh
epoberezkin: I think the point of having them here is that we can reliably rebuild the snapshot in case some of their files change…

scripts/smp-server-digitalocean-droplet/files/opt/simplex/on_login.sh
efim-poberezkin: But it's calculated in runtime, isn't it? Which means you'd have to dynamically edit the config file. Also, you can't supply the hash as a developer as part of configuration, and you cannot reuse config file after the hash has been added. I'm for storing the hash in a file, but not in the ini - let's just have a separate file where server writes hash on initialization, smth like `/etc/opt/simplex/pubkey_hash`?

scripts/smp-server-digitalocean-droplet/scripts/03-init-start.sh
npoberezkin: we were grepping it further, but I've already removed this line anyway, will commit soon

scripts/smp-server-digitalocean-droplet/files/opt/simplex/on_login.sh
npoberezkin: already done it, checking, commit soon

efim-poberezkin (approved):

PR #153, opened 2021-05-29
database migrations

epoberezkin:


src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -66,15 +70,42 @@ createSQLiteStore dbFilePath = do
...
+    MigrateError e -> confirmOrExit $ "Database is possibly corrupted: " <> e
+    ms@MigrateDown {} -> do
+      confirmOrExit
+        "The chat database has a newer version than the app.\n\
+        \It is recommended to run a newer version of the app.\n\
+        \If you continue the database will be backed up and downgraded."
+      migrate ms
+    ms@MigrateUp {dbMigrations} -> do
+      unless (null dbMigrations) $
+        confirmOrExit "The app has a newer version that the chat database - it will be backed up and upgraded."
```
efim-poberezkin, line 93: that -> than

src/Simplex/Messaging/Agent/Store/SQLite/Migrations/M_20210101_initial.hs
```diff
@@ -0,0 +1,142 @@
...
+import Database.SQLite.Simple.QQ (sql)
+import Simplex.Messaging.Agent.Store.SQLite.Migrations.Types (SchemaMigration (..))
+
+m_20210101_initial :: SchemaMigration
+m_20210101_initial =
+  SchemaMigration
+    { name = "20210101_initial",
+      up =
+        [sql|
+          CREATE TABLE IF NOT EXISTS servers(
```
efim-poberezkin, line 14: store these in from identically named .sql files? they're becoming hard to read w/t syntax highlighting

efim-poberezkin:
not a fan of these up and down migrations due to many reasons:
- have to write two way migrations - writing migrations becomes 2x more prone to error
- migration down deletes from `migrations` table
- more complex implementation of migration logic
- data is not being migrated down anyway (neither it does up, but this we'd have to support)

I think we should use migrations as a tool for us to upgrade schema one way, rather than for a user to allow having non matching versions of database and application (who'd want it anyway?) - when the mismatch happens due to application version upgrade the application would then read append only migration table and based on the new migrations it has play them out. List of migrations doesn't have to be duplicated in the code, but built from the files in migrations folder. The first migration would be creation of the migrations table

src/Simplex/Messaging/Agent/Store/SQLite/Migrations/M_20210101_initial.hs
epoberezkin: moved to .sql files

epoberezkin:
removed down migrations

> neither it does up, but this we'd have to support

up migrations have to migrate data, not only schema.

src/Simplex/Messaging/Agent/Store/SQLite.hs
epoberezkin: you can't back up the app, so unnecessary clarification

efim-poberezkin (approved):
Cool!

The only thing I question is exposing `Migrations.app` - this doesn't seem to have any benefit, we could just "call" `app` directly inside `get` w/t parameterizing it.

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -41,51 +42,70 @@ import Database.SQLite.Simple.ToField (ToField (..))
...
+
+migrateSchema :: SQLiteStore -> IO ()
+migrateSchema SQLiteStore {dbConn, dbFilePath, dbNew} = do
+  Migrations.initialize dbConn
+  Migrations.get dbConn Migrations.app >>= \case
+    Left e -> confirmOrExit $ "Database error: " <> e
+    Right [] -> pure ()
+    Right ms -> do
+      unless dbNew $ do
+        confirmOrExit "The app has a newer version than the database - it will be backed up and upgraded."
```
efim-poberezkin, line 86: and re-phrase this please - the second part now ambiguously reads as if it's the app that is backed up and upgraded (which is ofc total nonsense from the context).
either:
`The app has a newer version than the database - database will be backed up and upgraded.`
or
`The database has an older version than the app - it will be backed up and upgraded.`

epoberezkin: "it" can bind both the subject and the object, and it is usually thought of as binding the nearest, which is the database... Also you cannot backup the app anyway.

PR #154, opened 2021-05-30
broadcast commands

epoberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -271,9 +260,54 @@ processCommand c@AgentClient {sndQ} st (ATransmission corrId entity cmd) =
...
+  DEL -> withStore (deleteBcast st bId) >> ok
+  where
+    sendMsg :: MsgBody -> ConnAlias -> m ()
+    sendMsg msgBody cId = sendMessage c st corrId (Conn cId) msgBody
+
+    ok :: m ()
+    ok = respond bcast OK
+
+    respond :: EntityCommand t c' => Entity t -> ACommand 'Agent c' -> m ()
+    respond ent resp = atomically . writeTBQueue (sndQ c) $ ATransmission corrId ent resp
```
efim-poberezkin, line 309: seems like this `respond` and the one on line 263 can be merged into one `respondCommand`

src/Simplex/Messaging/Agent.hs
```diff
@@ -271,9 +260,54 @@ processCommand c@AgentClient {sndQ} st (ATransmission corrId entity cmd) =
...
+          SndMsgData {internalId, internalSndId, internalTs, msgBody, internalHash = msgHash}
+      sendAgentMessage c sq msgStr
+      atomically . writeTBQueue (sndQ c) $ ATransmission corrId (Conn cId) $ SENT (unId internalId)
+
+processBroadcastCommand ::
+  forall c m. (AgentMonad m, EntityCommand 'Broadcast_ c) => AgentClient -> SQLiteStore -> ACorrId -> Entity 'Broadcast_ -> ACommand 'Client c -> m ()
+processBroadcastCommand c st corrId bcast@(Broadcast bId) = \case
+  NEW -> withStore (createBcast st bId) >> ok
+  ADD (Conn cId) -> withStore (addBcastConn st bId cId) >> ok
+  REM (Conn cId) -> withStore (removeBcastConn st bId cId) >> ok
```
efim-poberezkin, line 297: it feels like using entity for connection here is wrong (`Conn`), since we refer to entity as smth that prepends command and to which the command is applied (bcast), and this is only a parameter for command. How about switching these to using `ConnAlias`? - It is even expressed then as `ConnAlias` in store

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -283,6 +294,10 @@ data StoreError
   | -- | Wrong connection type, e.g. "send" connection when "receive" or "duplex" is expected, or vice versa.
     -- 'upgradeRcvConnToDuplex' and 'upgradeSndConnToDuplex' do not allow duplex connections - they would also return this error.
     SEBadConnType ConnType
+  | -- | Broadcast ID not found
```
efim-poberezkin, line 297: dot

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -281,6 +283,48 @@ instance (MonadUnliftIO m, MonadError StoreError m) => MonadAgentStore SQLiteSto
   getMsg :: SQLiteStore -> ConnAlias -> InternalId -> m Msg
   getMsg _st _connAlias _id = throwError SENotImplemented
 
+  createBcast :: SQLiteStore -> BroadcastId -> m ()
+  createBcast SQLiteStore {dbConn} bId =
+    liftIOEither $
+      checkConstraint SEBcastDuplicate $
+        DB.execute dbConn "INSERT INTO broadcasts (broadcast_id) VALUES (?);" (Only bId)
```
efim-poberezkin, line 290: please let's stick to named parameters here and in other bcast methods. I understand it's only 1 here, but I'd rather be consistent with the rest of the code than save a few symbols

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -686,3 +730,31 @@ updateHashSnd_ dbConn connAlias SndMsgData {..} =
...
       ":last_internal_snd_msg_id" := internalSndId
     ]
+
+-- * Broadcast helpers
+
+checkBroadcast :: DB.Connection -> BroadcastId -> IO (Either StoreError a) -> IO (Either StoreError a)
+checkBroadcast dbConn bId action =
+  withTransaction dbConn $ do
+    ok <- bcastExists_ dbConn bId
+    if ok then action else pure $ Left SEBcastNotFound
```
efim-poberezkin, line 740: could we remove callback from the helper, please? - this can be moved to calling functions, it just reads better if the logic flow is flattened there

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -686,3 +730,31 @@ updateHashSnd_ dbConn connAlias SndMsgData {..} =
...
+
+bcastExists_ :: DB.Connection -> BroadcastId -> IO Bool
+bcastExists_ dbConn bId = not . null <$> queryBcast
+  where
+    queryBcast :: IO [Only BroadcastId]
+    queryBcast = DB.query dbConn "SELECT broadcast_id FROM broadcasts WHERE broadcast_id = ?" (Only bId)
```
efim-poberezkin, line 746: `;` in query

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -686,3 +730,31 @@ updateHashSnd_ dbConn connAlias SndMsgData {..} =
...
+bcastConnExists_ dbConn bId connAlias = not . null <$> queryBcastConn
+  where
+    queryBcastConn :: IO [(BroadcastId, ConnAlias)]
+    queryBcastConn =
+      DB.query
+        dbConn
+        [sql|
+          SELECT broadcast_id, conn_alias
+          FROM broadcast_connections
+          WHERE broadcast_id = ? AND conn_alias = ?
```
efim-poberezkin, line 758: same

src/Simplex/Messaging/Agent/Store/SQLite.hs
efim-poberezkin: or maybe it's okay.. not sure, kind of a grew a bit on me after several functions

src/Simplex/Messaging/Agent/Store.hs
epoberezkin: doesn't matter :)

src/Simplex/Messaging/Agent/Store/SQLite.hs
epoberezkin: I suggest we have up to three without naming them, as long as they as sufficiently close - it's equivalent to parameters in the functions. Otherwise the same parameter has to be named 3 times (2 times as query param and 1 as a name in code).

src/Simplex/Messaging/Agent/Store/SQLite.hs
epoberezkin: added but looks like it works without

src/Simplex/Messaging/Agent.hs
epoberezkin: I think it's better like that as it allows for future extension and also clarifies what it is.

src/Simplex/Messaging/Agent.hs
epoberezkin: it would need all two more parameters that are now in scope - easier to repeat it...

PR #156, opened 2021-06-01
introduction protocol

epoberezkin:


migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,26 @@
...
+  conn_info BLOB, -- info about another connection
+  conn_alias BLOB REFERENCES connections (conn_alias) -- created connection
+    ON DELETE CASCADE
+    DEFERRABLE INITIALLY DEFERRED,
+  invitation TEXT -- NULL if it's an initial introduction
+  status TEXT -- '', 'ACPT', 'CON'
+) WITHOUT ROWID;
+
+ALTER TABLE connections
+  ADD via_inv INTEGER REFERENCES conn_invitations (inv_id) ON DELETE RESTRICT;
```
efim-poberezkin, line 24: why is `via_inv` INTEGER but `inv_id` it references is BLOB?

migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,26 @@
...
+ALTER TABLE connections
+  ADD conn_level INTEGER DEFAULT 0;
```
efim-poberezkin, line 26: is it integrity level?

migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,26 @@
+CREATE TABLE IF NOT EXISTS conn_intros (
```
efim-poberezkin, line 1: now that we have migrations in place why do we need `IF NOT EXISTS`? it should not exist - if it exists at this point (when this migration hasn't been played out and registered in database) then the database is in potentially in inconsistent state / compromised / anything else - just fail with error and let the user figure out why he has this inconsistency in the first place

efim-poberezkin: shouldn't we at least also have `drop if exists` in this case then? By only having `create if not exists` we basically leave it to chance that if the table does exist it is coincidentally consistent with the rest of the schema and application - this leads to non idempotent migrations and opens up a way for cryptic errors in application. I just don't see any advantages to having only `create if not exists` except for possible minor conveniences during development

migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,26 @@
...
+  intro_id BLOB NOT NULL PRIMARY KEY,
+  to_conn BLOB REFERENCES connections (conn_alias) ON DELETE SET NULL,
```
efim-poberezkin, line 3: is introduction unique on `to_conn`, `from_conn`? if yes, should it be primary key? if not, do we allow and how do we process multiple invitations between same 2 connections?

migrations/20210602_introductions.sql
npoberezkin: I either do not see any case in which `if not exists` may be needed

migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,26 @@
...
+  re_info BLOB, -- info about "re" connection sent to "to" connection
+  re_status TEXT -- '', REQ, CON
+) WITHOUT ROWID;
+
+CREATE TABLE IF NOT EXISTS conn_invitations (
+  inv_id BLOB NOT NULL PRIMARY KEY,
+  via_conn BLOB REFERENCES connections (conn_alias) ON DELETE SET NULL,
+  ext_intro_id BLOB NOT NULL,
+  conn_info BLOB, -- info about another connection
+  conn_alias BLOB REFERENCES connections (conn_alias) -- created connection
```
efim-poberezkin, line 16: same - trying to figure out what is the real "unique" here.. is it just created connection `conn_alias`? can't it be made primary key?

migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,26 @@
...
+  to_status TEXT, -- '', INTRO, INV, CON
+  re_conn BLOB REFERENCES connections (conn_alias) ON DELETE SET NULL,
+  re_info BLOB, -- info about "re" connection sent to "to" connection
+  re_status TEXT -- '', REQ, CON
+) WITHOUT ROWID;
+
+CREATE TABLE IF NOT EXISTS conn_invitations (
+  inv_id BLOB NOT NULL PRIMARY KEY,
+  via_conn BLOB REFERENCES connections (conn_alias) ON DELETE SET NULL,
+  ext_intro_id BLOB NOT NULL,
```
efim-poberezkin, line 14: `ext` -> `external` for consistency

migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,26 @@
...
+  conn_info BLOB, -- info about another connection
+  conn_alias BLOB REFERENCES connections (conn_alias) -- created connection
+    ON DELETE CASCADE
+    DEFERRABLE INITIALLY DEFERRED,
+  invitation TEXT -- NULL if it's an initial introduction
```
efim-poberezkin, line 19: comma missing?

migrations/20210602_introductions.sql
efim-poberezkin: also please separate constraints below fields definition for consistency - we have it like this in the initial migration.
i.e.
```
PRIMARY KEY (inv_id),
FOREIGN KEY (conn_alias) REFERENCES ...
```

migrations/20210602_introductions.sql
epoberezkin: mistake - fixing

migrations/20210602_introductions.sql
epoberezkin: good question. I don't think should be - there may be a reason to send a repeat introduction, e.g. when the first one was not accepted. Introduction is identified by its ID that is used in messages to reference it.

migrations/20210602_introductions.sql
epoberezkin: This is a taste preference I disagree with. I prefer when constraints should be written on separate lines only when they are for multiple fields, I assumed that's what we had in initial schema.

epoberezkin: > what is the real "unique" here..

invitation ID that is referred to in commands

> is it just created connection conn_alias? can't it be made primary key?

no, because it is created in result of ACPT command - it is nullable (NULL initially)

migrations/20210602_introductions.sql
epoberezkin: updated

migrations/20210602_introductions.sql
epoberezkin: yes. why did it work I wonder? are commas optional?

migrations/20210602_introductions.sql
epoberezkin: this is connection "distance" - 0 is for direct, 1 is for "via direct", 2 is for "via 1" etc.

probably trustworthiness/security is a better word.

efim-poberezkin: yep, that's what I meant. `known_intermediaries`? :)

migrations/20210602_introductions.sql
efim-poberezkin: I'm starting to doubt but I'm pretty sure they aren't. maybe it was a different version that worked? w/e

migrations/20210602_introductions.sql
epoberezkin: they can be collected from invitations :)

migrations/20210602_introductions.sql
epoberezkin: possible...

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -248,83 +268,213 @@ instance (MonadUnliftIO m, MonadError StoreError m) => MonadAgentStore SQLiteSto
...
         <$> DB.query dbConn "SELECT conn_alias FROM broadcast_connections WHERE broadcast_id = ?;" (Only bId)
 
+  createIntro :: SQLiteStore -> TVar ChaChaDRG -> NewIntroduction -> m IntroId
+  createIntro SQLiteStore {dbConn} gVar NewIntroduction {toConn, reConn, reInfo} =
+    liftIOEither . createWithRandomId gVar $ \introId ->
+      DB.execute
+        dbConn
+        [sql|
+          INSERT INTO conn_intros
+          (intro_id, to_conn, re_conn, re_info) VALUES (?, ?, ?, ?);
```
efim-poberezkin, line 356: 1 or 2 is okay, but this will quickly spiral out of control. please switch to named parameters, they're just easier to verify than positional

efim-poberezkin: You even do named below in update with 4 params

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -248,83 +268,213 @@ instance (MonadUnliftIO m, MonadError StoreError m) => MonadAgentStore SQLiteSto
...
+
+  createInvitation :: SQLiteStore -> TVar ChaChaDRG -> NewInvitation -> m InvitationId
+  createInvitation SQLiteStore {dbConn} gVar NewInvitation {viaConn, externalIntroId, entityInfo, qInfo} =
+    liftIOEither . createWithRandomId gVar $ \invId ->
+      DB.execute
+        dbConn
+        [sql|
+          INSERT INTO conn_invitations
+          (inv_id, via_conn, external_intro_id, conn_info, queue_info)
+          VALUES (?, ?, ?, ?, ?);
```
efim-poberezkin, line 419: Same here

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -452,85 +619,99 @@ insertSndQueue_ dbConn SndQueue {..} = do
...
     dbConn
     [sql|
       INSERT INTO connections
-        ( conn_alias, rcv_host, rcv_port, rcv_id, snd_host, snd_port, snd_id,
+        ( conn_alias, rcv_host, rcv_port, rcv_id, snd_host, snd_port, snd_id, via_inv, conn_level,
           last_internal_msg_id, last_internal_rcv_msg_id, last_internal_snd_msg_id,
           last_external_snd_msg_id, last_rcv_msg_hash, last_snd_msg_hash)
       VALUES
-        (:conn_alias,     NULL,     NULL,   NULL,:snd_host,:snd_port,:snd_id,
+        (:conn_alias, NULL, NULL, NULL, :snd_host, :snd_port, :snd_id, :via_inv, :conn_level,
```
efim-poberezkin, line 640: These were easy to verify visually when they had the same horizontal indentation

tests/AgentTests/SQLiteTests.hs
```diff
@@ -132,12 +138,15 @@ testForeignKeysEnabled =
...
+cData1 :: ConnData
+cData1 = ConnData {connId = "conn1", viaInv = Nothing, connLevel = 1}
+
 rcvQueue1 :: RcvQueue
 rcvQueue1 =
   RcvQueue
     { server = SMPServer "smp.simplex.im" (Just "5223") testKeyHash,
       rcvId = "1234",
-      connAlias = "conn1",
+      -- connAlias = "conn1",
```
efim-poberezkin, line 149: Remove commented code

efim-poberezkin (approved):

migrations/20210602_introductions.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE conn_intros (
```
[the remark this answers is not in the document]
epoberezkin: can we cascade delete?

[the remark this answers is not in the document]
epoberezkin: changed to  ON DELETE CASCADE for conn_intros - no much point it surviving indeed, quite the opposite.

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -30,41 +35,52 @@ import qualified Simplex.Messaging.Protocol as SMP
...
 class Monad m => MonadAgentStore s m where
   -- Queue and Connection management
-  createRcvConn :: s -> RcvQueue -> m ()
-  createSndConn :: s -> SndQueue -> m ()
-  getConn :: s -> ConnAlias -> m SomeConn
-  getAllConnAliases :: s -> m [ConnAlias] -- TODO remove - hack for subscribing to all
+  createRcvConn :: s -> TVar ChaChaDRG -> ConnData -> RcvQueue -> m ConnId
+  createSndConn :: s -> TVar ChaChaDRG -> ConnData -> SndQueue -> m ConnId
+  getConn :: s -> ConnId -> m SomeConn
+  getAllConnAliases :: s -> m [ConnId] -- TODO remove - hack for subscribing to all
```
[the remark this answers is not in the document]
epoberezkin: changed

PR #158, opened 2021-06-12
remove entities, remove broadcasts

epoberezkin:


efim-poberezkin (approved):

efim-poberezkin:
do you want to keep respective rfcs for now, or shall we remove them as well?

PR #163, opened 2021-06-24
ask client for confirmation of sender; make establishment of connection asynchronous

efim-poberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -434,17 +469,18 @@ processSMPTransmission c@AgentClient {subQ} (srv, rId, cmd) = do
...
-              withStore $ \st -> setRcvQueueStatus st rq Confirmed
-              -- TODO update sender key in the store?
-              secureQueue c rq senderKey
-              withStore $ \st -> setRcvQueueStatus st rq Secured
+            New -> case cType of
+              SCRcv -> do
+                g <- asks idsDrg
+                let newConfirmation = NewConfirmation {connId, senderKey, senderConnInfo = cInfo}
+                confirmationId <- withStore $ \st -> createConfirmation st g newConfirmation
+                notify $ CNFRM confirmationId cInfo
```
epoberezkin, line 481: CONF

src/Simplex/Messaging/Agent.hs
```diff
@@ -280,6 +293,27 @@ joinConn c connId qInfo (ReplyMode replyMode) viaInv connLevel = do
       withStore $ \st -> upgradeSndConnToDuplex st cId rq
       sendControlMessage c sq $ REPLY qInfo'
 
+letConf :: AgentMonad m => AgentClient -> ConnId -> ConfirmationId -> ConnInfo -> m ConnId
+letConf c connId confirmationId ownConnInfo =
+  withStore (`getConn` connId) >>= \case
+    SomeConn SCDuplex (DuplexConnection _ rq _) -> processConfirmation' rq
```
epoberezkin, line 299: should not be allowed for DuplexConnection

src/Simplex/Messaging/Agent.hs
```diff
@@ -280,6 +293,27 @@ joinConn c connId qInfo (ReplyMode replyMode) viaInv connLevel = do
...
+    SomeConn SCRcv (RcvConnection _ rq) -> processConfirmation' rq
+    _ -> throwError $ CONN SIMPLEX
```
epoberezkin, line 301: better use CONN NOT_FOUND or CMD PROHIBITED

src/Simplex/Messaging/Agent.hs
```diff
@@ -262,14 +275,14 @@ newConn c connId viaInv connLevel = do
...
+joinConn :: forall m. AgentMonad m => AgentClient -> ConnId -> SMPQueueInfo -> ConnInfo -> Maybe InvitationId -> Int -> m ConnId
+joinConn c connId qInfo cInfo viaInv connLevel = do
   (sq, senderKey, verifyKey) <- newSendQueue qInfo
   g <- asks idsDrg
   let cData = ConnData {connId, viaInv, connLevel}
   connId' <- withStore $ \st -> createSndConn st g cData sq
-  connectToSendQueue c sq senderKey verifyKey
-  when (replyMode == On) $ createReplyQueue connId' sq
+  connectToSendQueue c sq senderKey verifyKey cInfo
+  createReplyQueue connId' sq
```
epoberezkin, line 285: createReplyQueue should be called after HELLO successfully sent, that is in activateQueue

epoberezkin (changes requested):

src/Simplex/Messaging/Agent.hs
```diff
@@ -39,6 +39,7 @@ module Simplex.Messaging.Agent
     getSMPAgentClient,
     createConnection,
     joinConnection,
+    letConfirmation,
```
epoberezkin, line 42: confirmConnection would be more consistent

src/Simplex/Messaging/Agent.hs
```diff
@@ -255,31 +266,58 @@ processCommand c (connId, cmd) = case cmd of
...
   connId' <- withStore $ \st -> createSndConn st g cData sq
-  connectToSendQueue c sq senderKey verifyKey
-  when (replyMode == On) $ createReplyQueue connId' sq
+  confirmQueue c sq senderKey cInfo
+  activateQueueJoining c connId' sq verifyKey 1 60
   pure connId'
+
+activateQueueJoining :: forall m. AgentMonad m => AgentClient -> ConnId -> SndQueue -> VerificationKey -> Int -> Int -> m ()
+activateQueueJoining c connId sndQ verifyKey initialDelaySec increaseDelayAfterSec = do
+  let onActivationCallback = removeActivation c connId >> cleanupSignatureKey sndQ >> createReplyQueue connId sndQ
```
epoberezkin, line 288: it would be better to put it into where

src/Simplex/Messaging/Agent.hs
```diff
@@ -255,31 +266,58 @@ processCommand c (connId, cmd) = case cmd of
...
+letConf :: AgentMonad m => AgentClient -> ConnId -> ConfirmationId -> ConnInfo -> m ConnId
+letConf c connId confirmationId ownConnInfo =
+  withStore (`getConn` connId) >>= \case
+    SomeConn SCRcv (RcvConnection _ rq) -> processConfirmation' rq
+    _ -> throwError $ CMD PROHIBITED
+  where
+    processConfirmation' :: AgentMonad m => RcvQueue -> m ConnId
+    processConfirmation' rq = do
+      withStore $ \st -> approveConfirmation st confirmationId ownConnInfo
+      confirmation <- withStore $ \st -> getConfirmation st confirmationId
```
epoberezkin, line 309: these two methods can be probably merged into one?

src/Simplex/Messaging/Agent.hs
epoberezkin: also, why removeActivation is needed in callback - it should happen in all cases, so it can be inside activation itself, no?

src/Simplex/Messaging/Agent.hs
```diff
@@ -255,31 +266,58 @@ processCommand c (connId, cmd) = case cmd of
...
+      let sndKey = senderKey (confirmation :: Confirmation)
+      processConfirmation c rq sndKey
+      withStore $ \st -> removeApprovedConfirmation st connId
+      pure connId
+
+processConfirmation :: AgentMonad m => AgentClient -> RcvQueue -> SenderPublicKey -> m ()
+processConfirmation c rq sndKey = do
+  withStore $ \st -> setRcvQueueStatus st rq Confirmed
```
epoberezkin, line 317: the queue should be marked as confirmed when SMP confirmation arrives, not when it is LET by the initiating party

epoberezkin: although, given that strictly speaking there can be several confirmation, maybe it's ok...

src/Simplex/Messaging/Agent.hs
```diff
@@ -312,9 +351,26 @@ acceptInv c connId invId connInfo =
 subscribeConnection' :: AgentMonad m => AgentClient -> ConnId -> m ()
 subscribeConnection' c connId =
   withStore (`getConn` connId) >>= \case
-    SomeConn _ (DuplexConnection _ rq _) -> subscribeQueue c rq connId
+    SomeConn _ (DuplexConnection _ rq sq) -> case status (sq :: SndQueue) of
+      Confirmed -> do
+        let verifyKey = C.publicKey $ sndPrivateKey sq
+        case verifyKey of
+          Nothing -> throwError $ CONN NO_PUBLIC_KEY
```
epoberezkin, line 358: I would simply return INTERNAL here, it indicates a logical implementation error, not a protocol error

src/Simplex/Messaging/Agent.hs
```diff
@@ -312,9 +351,26 @@ acceptInv c connId invId connInfo =
...
+          Just vk -> do
+            activateQueueInitiating c connId sq vk 5 0
+            subscribeQueue c rq connId
+      Active -> subscribeQueue c rq connId
+      _ -> throwError $ CONN BAD_QUEUE_STATUS
```
epoberezkin, line 363: ditto here

epoberezkin: also, we don't generally validate that the queue is in the expected status, so it could be just `_ -> subscribeQueue c rq connId`

src/Simplex/Messaging/Agent.hs
```diff
@@ -312,9 +351,26 @@ acceptInv c connId invId connInfo =
...
+    SomeConn _ (SndConnection _ sq) -> case status (sq :: SndQueue) of
+      Confirmed -> do
+        let verifyKey = C.publicKey $ sndPrivateKey sq
+        case verifyKey of
+          Nothing -> throwError $ CONN NO_PUBLIC_KEY
```
epoberezkin, line 368: same

src/Simplex/Messaging/Agent.hs
```diff
@@ -312,9 +351,26 @@ acceptInv c connId invId connInfo =
...
+          Just vk -> do
+            activateQueueJoining c connId sq vk 5 0
+      Active -> throwError $ CONN SIMPLEX
+      _ -> throwError $ CONN BAD_QUEUE_STATUS
```
epoberezkin, line 372: same

migrations/20210624_confirmations.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE conn_confirmations (
+  confirmation_id BLOB NOT NULL,
+  conn_alias BLOB NOT NULL,
+  sender_key BLOB NOT NULL,
+  sender_conn_info BLOB NOT NULL,
+  approved INTEGER NOT NULL,
+  own_conn_info BLOB,
+  created_at TEXT NOT NULL DEFAULT (datetime('now')),
+  PRIMARY KEY (confirmation_id),
+  FOREIGN KEY (conn_alias)
```
epoberezkin, line 10: Suggest we stop separating field indexes from fields, when possible, it only increases cognitive load, doesn't solve any problem.

src/Simplex/Messaging/Agent.hs
```diff
@@ -407,8 +463,8 @@ processSMPTransmission c@AgentClient {subQ} (srv, rId, cmd) = do
           msg <- decryptAndVerify rq msgBody
           let msgHash = C.sha256Hash msg
           case parseSMPMessage msg of
-            Left e -> notify $ ERR e
-            Right (SMPConfirmation senderKey) -> smpConfirmation senderKey
+            Left e -> notify c connId $ ERR e
```
epoberezkin, line 466: rather than changing all calls to local notify for the sake of taking it out, it would be better to just use another function where it was needed (or just its implementation - it's one line).

The whole point of having it was not to pass these parameters at call sites.

src/Simplex/Messaging/Agent.hs
```diff
@@ -573,16 +624,43 @@ processSMPTransmission c@AgentClient {subQ} (srv, rId, cmd) = do
...
+confirmQueue c sq senderKey cInfo = do
+  sendConfirmation c sq senderKey cInfo
   withStore $ \st -> setSndQueueStatus st sq Confirmed
-  sendHello c sq verifyKey
-  withStore $ \st -> setSndQueueStatus st sq Active
 
-newSendQueue ::
+activateQueueInitiating :: AgentMonad m => AgentClient -> ConnId -> SndQueue -> VerificationKey -> Int -> Int -> m ()
+activateQueueInitiating c connId sndQ verifyKey initialDelaySec increaseDelayAfterSec = do
+  let onActivationCallback = removeActivation c connId >> cleanupSignatureKey sndQ >> connected c connId
```
epoberezkin, line 634: better in where

epoberezkin: also removeActivation is common to all cases

epoberezkin: `cleanupSignatureKey sndQ` also seems to be common

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -80,8 +84,9 @@ newAgentClient agentEnv = do
   smpClients <- newTVar M.empty
   subscrSrvrs <- newTVar M.empty
   subscrConns <- newTVar M.empty
+  pendingActivations <- newTVar M.empty
```
epoberezkin, line 87: just activations?

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -126,9 +131,17 @@ getSMPServerClient c@AgentClient {smpClients, msgQ} srv =
     notifySub :: ConnId -> IO ()
     notifySub connId = atomically $ writeTBQueue (subQ c) ("", connId, END)
 
+cleanupAgentClient :: MonadUnliftIO m => AgentClient -> m ()
```
epoberezkin, line 134: "close" is probably better word here...

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -213,6 +225,16 @@ removeSubscription AgentClient {subscrConns, subscrSrvrs} connId = atomically $
       let cs' = S.delete connId cs
        in if S.null cs' then Nothing else Just cs'
 
+addActivation :: MonadUnliftIO m => AgentClient -> ConnId -> Async () -> m ()
+addActivation c connId a =
+  atomically $
+    modifyTVar (pendingActivations c) $ M.insert connId a
```
epoberezkin, line 231: no need to have them on 3 separate lines - it's short...

epoberezkin: also, it's more common to define such functions in STM monad and use atomically at call sites

epoberezkin: also - there should probably be check to prevent adding activation for the same connection - SUB can be used multiple times, currently it'll spawn activation every time SUB is used?

src/Simplex/Messaging/Agent.hs
```diff
@@ -573,16 +624,43 @@ processSMPTransmission c@AgentClient {subQ} (srv, rId, cmd) = do
...
+confirmQueue :: AgentMonad m => AgentClient -> SndQueue -> SenderPublicKey -> ConnInfo -> m ()
+confirmQueue c sq senderKey cInfo = do
+  sendConfirmation c sq senderKey cInfo
   withStore $ \st -> setSndQueueStatus st sq Confirmed
-  sendHello c sq verifyKey
-  withStore $ \st -> setSndQueueStatus st sq Active
 
-newSendQueue ::
+activateQueueInitiating :: AgentMonad m => AgentClient -> ConnId -> SndQueue -> VerificationKey -> Int -> Int -> m ()
+activateQueueInitiating c connId sndQ verifyKey initialDelaySec increaseDelayAfterSec = do
```
epoberezkin, line 633: why do we need initialDelaySec and increaseDelayAfterSec parameters and not have inside sendHello? The are always set to the same values

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -189,7 +191,12 @@ deriving instance Show (ACommand p)
 data SMPMessage
   = -- | SMP confirmation
     -- (see <https://github.com/simplex-chat/simplexmq/blob/master/protocol/simplex-messaging.md#send-message SMP protocol>)
-    SMPConfirmation SenderPublicKey
+    SMPConfirmation
+      { -- | sender's public key to use for authentication of sender's commands at the recepient's server
+        senderKey :: SenderPublicKey,
+        -- | sender's information to be associated with the connection, e.g. sender's profile information
+        cInfo :: ConnInfo
```
epoberezkin, line 198: cInfo -> connInfo

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -46,14 +46,20 @@ class Monad m => MonadAgentStore s m where
   setRcvQueueStatus :: s -> RcvQueue -> QueueStatus -> m ()
   setRcvQueueActive :: s -> RcvQueue -> VerificationKey -> m ()
   setSndQueueStatus :: s -> SndQueue -> QueueStatus -> m ()
+  updateSignatureKey :: s -> SndQueue -> SignatureKey -> m ()
+
+  -- Confirmations
+  createConfirmation :: s -> TVar ChaChaDRG -> NewConfirmation -> m ConfirmationId
+  getConfirmation :: s -> ConfirmationId -> m Confirmation
+  approveConfirmation :: s -> ConfirmationId -> ConnInfo -> m ()
```
epoberezkin, line 54: I don't see any situation where getConfirmation and approveConfirmation are used separately, so it could be just approveConfirmation that returns the confirmation.

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -46,14 +46,20 @@ class Monad m => MonadAgentStore s m where
...
+  getApprovedConfirmation :: s -> ConnId -> m Confirmation
+  removeApprovedConfirmation :: s -> ConnId -> m ()
```
epoberezkin, line 56: this doesn't seem used at all - probably if we want to remove confirmation after it's accepted we can just remove all of them?

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -429,6 +430,10 @@ data ConnectionErrorType
     DUPLICATE
   | -- | connection is simplex, but operation requires another queue
     SIMPLEX
+  | -- | queue status is prohibited in this context
+    BAD_QUEUE_STATUS
+  | -- | no public key stored for queue activation
+    NO_PUBLIC_KEY
```
epoberezkin, line 436: I don't think either of these errors are needed - they are implementation rather than protocol error, so should be reported as INTERNAL

migrations/20210624_confirmations.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE conn_confirmations (
+  confirmation_id BLOB NOT NULL,
+  conn_alias BLOB NOT NULL,
+  sender_key BLOB NOT NULL,
+  sender_conn_info BLOB NOT NULL,
+  approved INTEGER NOT NULL,
```
epoberezkin, line 6: DEFAULT 0

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -223,20 +245,20 @@ showServer srv = B.pack $ host srv <> maybe "" (":" <>) (port srv)
 logSecret :: ByteString -> ByteString
 logSecret bs = encode $ B.take 3 bs
 
-sendConfirmation :: forall m. AgentMonad m => AgentClient -> SndQueue -> SenderPublicKey -> m ()
-sendConfirmation c sq@SndQueue {server, sndId} senderKey =
+sendConfirmation :: forall m. AgentMonad m => AgentClient -> SndQueue -> SenderPublicKey -> ConnInfo -> m ()
+sendConfirmation c sq@SndQueue {server, sndId} senderKey cInfo =
```
epoberezkin, line 249: what I don't understand is at what point accepting party see the connection info of the initiating party in case we pass it back via SMP confirmation?

src/Simplex/Messaging/Agent.hs
efim-poberezkin: will rename to `allow`

src/Simplex/Messaging/Agent.hs
efim-poberezkin: will unite under transaction

src/Simplex/Messaging/Agent.hs
efim-poberezkin: also line 528 - Confirmation -> ApprovedConfirmation w/t Maybe

src/Simplex/Messaging/Agent/Client.hs
efim-poberezkin: TODO - check map and don't spawn thread if it exists

src/Simplex/Messaging/Agent.hs
efim-poberezkin: TODO type delay timing

src/Simplex/Messaging/Agent/Store.hs
efim-poberezkin: TODO remove all

src/Simplex/Messaging/Agent/Protocol.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent/Store.hs
epoberezkin: done

src/Simplex/Messaging/Agent/Store.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: done

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: done

src/Simplex/Messaging/Agent.hs
epoberezkin: done

src/Simplex/Messaging/Agent/Protocol.hs
epoberezkin: done

PR #174, opened 2021-07-26
decrease initial delay for HELLO retries on online activation

efim-poberezkin:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -285,7 +285,7 @@ sendHello c sq@SndQueue {server, sndId, sndPrivateKey} verifyKey RetryInterval {
           let newDelay =
                 if elapsedTime < increaseAfter || delay == maxInterval
                   then delay
-                  else min (delay * 3 `div` 2) maxInterval
+                  else min (delay * 20 `div` 19) maxInterval
```
epoberezkin, line 288: why the multiple is so close to 1 now? I think it would hammer it too much in async case.

src/Simplex/Messaging/Agent.hs
```diff
@@ -270,21 +270,24 @@ newConn c connId viaInv connLevel = do
...
-minute = 60_000_000
+minute = 60 * sec
 
 onlineInterval :: RetryInterval
 onlineInterval =
   RetryInterval
-    { initialInterval = 1_000_000,
-      increaseAfter = minute,
+    { initialInterval = sec `div` 10,
+      increaseAfter = 2 * sec,
```
epoberezkin, line 283: I think it should be 0 and the multiple can remain 1.5

PR #203, opened 2021-11-20
TLS 1.3 transport

epoberezkin:


apps/smp-server/Main.hs
```diff
@@ -70,9 +73,16 @@ iniFile = combine cfgDir "smp-server.ini"
 defaultKeyFile :: FilePath
 defaultKeyFile = combine cfgDir "server_key"
 
+defaultPrivateKeyFile :: FilePath
+defaultPrivateKeyFile = combine cfgDir "localhost.key"
```
epoberezkin, line 77:
```suggestion
defaultPrivateKeyFile = combine cfgDir "server.key"
```

apps/smp-server/Main.hs
```diff
@@ -70,9 +73,16 @@ iniFile = combine cfgDir "smp-server.ini"
...
+
+defaultCertificateFile :: FilePath
+defaultCertificateFile = combine cfgDir "localhost.crt"
```
epoberezkin, line 80:
```suggestion
defaultCertificateFile = combine cfgDir "server.crt"
```

apps/smp-server/Main.hs
```diff
@@ -94,6 +104,11 @@ main = do
     ServerDelete -> do
       deleteServer
       putStrLn "Server key, config file and store log deleted"
+  where
+    checkPubkeyAlgorihtm :: String -> IO ()
+    checkPubkeyAlgorihtm alg
+      | alg == "ED448" || alg == "ED25519" = pure ()
+      | otherwise = error "unsupported public-key algorithm"
```
epoberezkin, line 111: there is a better function for exit we use other than error that simply crashes

```suggestion
      | otherwise = putStrLn ("unsupported public-key algorithm " <> alg) >> exitFailure
```

src/Simplex/Messaging/Transport.hs
```diff
@@ -120,26 +124,32 @@ data TProxy c = TProxy
...
     closeServer :: TVar (Set ThreadId) -> Socket -> IO ()
     closeServer clients sock = do
       readTVarIO clients >>= mapM_ killThread
       close sock
       void . atomically $ tryPutTMVar started False
     acceptConnection :: Transport c => Socket -> IO c
-    acceptConnection sock = accept sock >>= getServerConnection . fst
+    acceptConnection sock = do
+      (newSock, _) <- accept sock
+      let serverParams = mkServerParams credential
```
epoberezkin, line 151: can't it be in Env? why is it new for each client connection?

src/Simplex/Messaging/Transport.hs
```diff
@@ -182,28 +192,99 @@ startTCPClient host port = withSocketsDo $ resolve >>= tryOpen err
...
+    getLock <- newTMVarIO ()
+    getPartyConnection TLS {tlsContext, buffer, getLock}
+      `E.catch` \(e :: E.SomeException) -> putStrLn (party <> " exception: " <> show e) >> E.throwIO e
+
+closeTLS :: T.Context -> IO ()
+closeTLS ctx =
+  (T.bye ctx >> T.contextClose ctx) -- sometimes socket was closed before 'TLS.bye'
+    `E.catch` (\(_ :: E.SomeException) -> pure ()) -- so we catch the 'Broken pipe' error here
+
+mkServerParams :: T.Credential -> T.ServerParams
```
epoberezkin, line 215: you could rename this one to tlsServerParams, export and use in Env instead of serverCredential

I think you suggested it initially, not sure why we decided this way?

src/Simplex/Messaging/Transport.hs
```diff
@@ -182,28 +192,99 @@ startTCPClient host port = withSocketsDo $ resolve >>= tryOpen err
...
+  def
+    { T.serverWantClientCert = False,
+      T.serverShared = def {T.sharedCredentials = T.Credentials [credential]},
+      T.serverHooks = def,
+      T.serverSupported = supportedParameters
+    }
+
+clientParams :: T.ClientParams
+clientParams =
+  (T.defaultParamsClient "localhost" "5223")
```
epoberezkin, line 226: I don't think this would work with remote server, I think it should be actual host and port

src/Simplex/Messaging/Transport.hs
```diff
@@ -182,28 +192,99 @@ startTCPClient host port = withSocketsDo $ resolve >>= tryOpen err
...
+    { T.clientShared = def,
+      T.clientHooks = def {T.onServerCertificate = \_ _ _ _ -> pure []},
```
epoberezkin, line 228: this will probably change once you start validating hash

efim-poberezkin: yes

src/Simplex/Messaging/Transport.hs
epoberezkin: you could rename this one to tlsServerParams, export and use in Env instead of serverCredential

I think you suggested it initially, not sure why we decided this way?

src/Simplex/Messaging/Transport.hs
efim-poberezkin: I don't think I suggested exactly that, but yes we could do it this way, will change

src/Simplex/Messaging/Agent/Env/SQLite.hs
```diff
@@ -78,10 +86,16 @@ newSMPAgentEnv cfg = do
...
     -- 1st rsaKeySize is used by the RSA signature in each command,
     -- 2nd - by encrypted message body header
     -- 3rd - by message signature
     -- smpCommandSize - is the estimated max size for SMP command, queueId, corrId
     reservedMsgSize = 3 * rsaKeySize cfg + smpCommandSize (smpCfg cfg)
+    loadAgentCredential :: (MonadUnliftIO m') => AgentConfig -> m' T.Credential
+    loadAgentCredential AgentConfig {agentPrivateKeyFile, agentCertificateFile} =
+      liftIO (T.credentialLoadX509 agentCertificateFile agentPrivateKeyFile) >>= \case
+        Right cert -> pure cert
+        Left _ -> error "invalid credential"
```
epoberezkin, line 101: Please use exitFailure.
There is never a reason to use error outside of tests.

src/Simplex/Messaging/Server/Env/STM.hs
```diff
@@ -114,3 +120,9 @@ newEnv config = do
     addNotifier q = case notifier q of
       Nothing -> id
       Just (nId, _) -> M.insert nId (recipientId q)
+    loadServerCredential :: ServerConfig -> m T.Credential
+    loadServerCredential ServerConfig {serverPrivateKeyFile, serverCertificateFile} =
+      -- TODO non lazy
+      liftIO (T.credentialLoadX509 serverCertificateFile serverPrivateKeyFile) >>= \case
+        Right cert -> pure cert
+        Left _ -> error "invalid credential"
```
epoberezkin, line 128: same here

efim-poberezkin (approved):

PR #210, opened 2021-11-28
URI syntax for SMP queues and connection requests

epoberezkin:


src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -290,19 +298,74 @@ smpServerP = SMPServer <$> server <*> optional port <*> optional kHash
...
+  sch <> "/" <> act <> "#/" <> queryStr
+  where
+    sch = case scheme of
+      CRSSimplex -> "simplex:"
+      CRSAppServer host port -> B.pack $ "https://" <> host <> maybe "" (':' :) port
+    act = case action of
+      CRAConnect -> "connect"
+    queryStr = renderSimpleQuery True [("smp", queues), ("e2e", key)]
+    queues = B.intercalate "," . map serializeSMPQueueUri $ L.toList smpQueues
+    key = C.serializePubKey encryptionKey
```
efim-poberezkin, line 328: we weren't sure whether it should be one key for multiple queues or a key per queue, do you have new considerations on this?

efim-poberezkin (approved):

src/Simplex/Messaging/Agent/Protocol.hs
epoberezkin: I think it should be managed/decided independently from queues, on the connection level. So if we decide to have multiple keys per connection (and it may as well be one per queue, but it doesn't make it a property of the queue in SMP protocol), we would change ConnectionRequest type. The main objective of this PR is to decouple E2E encryption from SMP protocol and from queue URIs, so we don't need to change the invitation syntax once we add support for multiple queues in the connection.

PR #244, opened 2021-12-29
update DB fields presentations: port, QueueStatus, RcvMsgStatus, SndMsgStatus

epoberezkin:


src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -257,13 +260,9 @@ data RcvMsg = RcvMsg
     brokerId :: BrokerId,
     brokerTs :: BrokerTs,
     rcvMsgStatus :: RcvMsgStatus,
-    -- | Timestamp of acknowledgement to broker, corresponds to `AcknowledgedToBroker` status.
+    -- | Timestamp of acknowledgement to broker, corresponds to `Acknowledged` status.
     -- Do not mix up with `brokerTs` - timestamp created at broker after it receives the message from sender.
```
efim-poberezkin, line 264:
```suggestion
    -- Don't confuse with `brokerTs` - timestamp created at broker after it receives the message from sender.
```

efim-poberezkin (approved):

PR #317, opened 2022-02-14
remove redundant asynchronous connection code

efim-poberezkin:


epoberezkin (approved):

epoberezkin:
This logic was redundant as it was trying to address a very low probability event (network failure in between operations, without any user interaction) in incorrect way.

PR #355, opened 2022-04-19
message TTL and periodic expiration; setting to prevent creation of the new queues

epoberezkin:


tests/SMPClient.hs
```diff
@@ -61,6 +61,9 @@ cfg =
       queueIdBytes = 24,
       msgIdBytes = 24,
       storeLog = Nothing,
+      allowNewQueues = True,
+      messageTTL = Just $ 7 * 86400, -- seconds, 7 days
+      expireMessagesInterval = Just 21600_000000, -- microseconds, 6 hours
```
jr-simplex, line 66: units are different here and in apps/smp-server/Main.hs

epoberezkin: fixed

jr-simplex (approved):

PR #371, opened 2022-05-11
server: CLI warning when deleting server configuration

epoberezkin:


src/Simplex/Messaging/Server/CLI.hs
```diff
@@ -58,11 +58,22 @@ protocolServerCLI cliCfg@ServerCLIConfig {iniFile, executableName} server =
       doesFileExist iniFile >>= \case
         True -> readIniFile iniFile >>= either exitError (runServer cliCfg server)
         _ -> exitError $ "Error: server is not initialized (" <> iniFile <> " does not exist).\nRun `" <> executableName <> " init`."
-    Delete -> cleanup cliCfg >> putStrLn "Deleted configuration and log files"
+    Delete -> do
+      confirmOrExit "WARNING: re-initializing the server will make all queues inaccessible, because the server identity (certificate fingerprint) will change.\nTHIS CANNOT BE UNDONE!"
```
jr-simplex, line 62:
```suggestion
      confirmOrExit "WARNING: deleting the server will make all queues inaccessible, because the server identity (certificate fingerprint) will change.\nTHIS CANNOT BE UNDONE!"
```

PR #390, opened 2022-06-07
SMP connection handshake v2

epoberezkin:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -142,15 +142,15 @@ type AgentMonad m = (MonadUnliftIO m, MonadReader Env m, MonadError AgentErrorTy
 
 class ProtocolServerClient msg where
   getProtocolServerClient :: AgentMonad m => AgentClient -> ProtocolServer -> m (ProtocolClient msg)
-  protocolError :: ErrorType -> AgentErrorType
+  clientProtocolError :: ErrorType -> AgentErrorType
```
epoberezkin, line 145: renamed, because the same name was used for another function in another type class

PR #421, opened 2022-06-23
ntf: notification server SDEL command; subscription deletion in agent

jr-simplex:


src/Simplex/Messaging/Notifications/Client.hs
```diff
@@ -162,15 +162,18 @@ instance FromField NtfSubAction where fromField = blobFieldDecoder smpDecode
 instance ToField NtfSubAction where toField = toField . smpEncode
 
 data NtfSubSMPAction
-  = NSAKey
+  = NSSmpAKey
+  | NSSmpADelete
```
epoberezkin, line 166:
```suggestion
  = NSANKey
  | NSANDelete
```

epoberezkin: or at least `NSASmpDelete`, `NSASmpKey`

PR #444, opened 2022-06-30
ntf: separate types into their own module

jr-simplex:


src/Simplex/Messaging/Notifications/Types.hs
```diff
@@ -0,0 +1,215 @@
...
+newNtfSubscription :: ConnId -> SMPServer -> Maybe NotifierId -> NtfServer -> NtfAgentSubStatus -> NtfSubscription
+newNtfSubscription connId smpServer ntfQueueId ntfServer ntfSubStatus =
+  NtfSubscription
+    { connId,
+      smpServer,
+      ntfQueueId,
+      ntfServer,
+      ntfSubId = Nothing,
+      ntfSubStatus
+    }
```
epoberezkin, line 215:
```suggestion
    }

```

PR #561, opened 2022-11-11
server: SMP basic auth

epoberezkin:


rfcs/2022-11-11-smp-basic-auth.md
```diff
@@ -0,0 +1,27 @@
...
+
+Single access password that can be optionally included in server address that is passed to app configuration. It will not be allowed in the existing contexts (and parsing will fail), to avoid accidentally leaking it. Server address with password will look this way: `smp://fingerprint:password@hosts`
+
+## Implementation plan
+
+1. A separate type to include server and password, so it can only be used where allowed.
+
+2. Server password to create queues will be configured in TRANSPORT section of INI file, as `create_password` parameter.
+
+3. The password will only be required in server configuration/address to create queues only, it won't be required for other receiving queue operations on already existing queues.
```
jr-simplex, line 17: I'm not saying it's wrong, but is there a reason to not require password for other operations? To not disrupt existing connections for users how did create queues on others' servers?

rfcs/2022-11-11-smp-basic-auth.md
```diff
@@ -0,0 +1,27 @@
...
+
+  - as a parameter of NEW command. Pros: a local change, that only needs checking when queue is created. Cons: protocol version change.
+  - as a separate command AUTH. Pros: allows to include additional parameters and potentially be extended beyond basic auth. Cons: more complex to manage server state, can be more difficult syntax in the future, if extended.
+  - as part of handshake (we currently ignore the unparsed part of handshake block, so it can be extended). Pros: probably, the simplest, and independent of the commands protocol – establishes create permission for the current session. Cons: the client won't know about whether it is able to create the queue until it tries (same as in case 1).
+
+  My preference is the last option. As a variant of the last option, we can add a server response/message that includes permission to create queues - it will only be sent to the clients who pass credential in handshake - that might simplify testing server connection (we currently do not do it). It might be unnecessary, as we could simply create and delete queue in case credential is passed as part of testing connection (and even sending a message to it).
```
jr-simplex, line 27: > ... we could simply create and delete queue in case credential is passed as part of testing connection (and even sending a message to it).

even though we do treat queues as inexpensive, this still seems wasteful..

jr-simplex: we don't do this now, I don't think there's greatly more reason to do it for authenticated connection if the implementation is correct.. client do retries or report error so it's also not that big of a problem - but if to solve it better via confirmation from server in the initial handshake / answer to AUTH

rfcs/2022-11-11-smp-basic-auth.md
epoberezkin: Yes

rfcs/2022-11-11-smp-basic-auth.md
epoberezkin: I think there is actually a value to perform the actual test of some basic operations, as part of agent “test server” function - more than from simply informing the client that it works

epoberezkin: And as result tell the user which operations succeeded and which failed, and then delete the queue as part of the test

epoberezkin: It can be used to pre-flight servers before changing config, because if you change config and some server doesn’t work you may permanently fail some things - like receiving files etc.

src/Simplex/Messaging/Protocol.hs
```diff
@@ -667,20 +670,29 @@ instance ProtocolTypeI p => Encoding (ProtocolServer p) where
...
-    TransportHosts host <- strP
-    port <- portP <|> pure ""
-    pure ProtocolServer {scheme, host, port, keyHash}
-    where
-      portP = show <$> (A.char ':' *> (A.decimal :: Parser Int))
+    strEncodeServer scheme (strEncode host) port keyHash Nothing
+  strP =
+    serverStrP >>= \case
+      (srv, Nothing) -> pure srv
+      _ -> fail "ProtocolServer with basic auth not allowed"
```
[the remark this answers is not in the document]
epoberezkin: :)

rfcs/2022-11-11-smp-basic-auth.md
jr-simplex: it's a good thing to do on config on change, but probably not on each new session

apps/smp-server/Main.hs
```diff
@@ -62,9 +63,18 @@ smpServerCLIConfig =
...
+              <> "log_stats: off\n\n"
               <> "[TRANSPORT]\n"
-              <> ("port: " <> defaultServerPort <> "\n")
+              <> ("port: " <> defaultServerPort <> "\n\n")
+              <> "# Set new_queues option to off to completely prohibit creating new messaging queues.\n"
+              <> "# This can be useful when you want to decommission the server, but not all connections are switched yet.\n"
+              <> "new_queues: on\n\n"
+              <> "# Use create_password option to enable basic auth to create new messaging queues.\n"
+              <> "# The password should be used as part of server address in client configuration:\n"
+              <> "# smp://fingerpring:password@host1,host2\n"
```
jr-simplex, line 74:
```suggestion
              <> "# smp://fingerprint:password@host1,host2\n"
```

apps/smp-server/Main.hs
```diff
@@ -92,7 +102,14 @@ smpServerCLIConfig =
...
                             -- if the setting is not set, it is enabled when store log is enabled
                             _ -> storeLogFile $> messagesPath,
-                    allowNewQueues = True,
+                    allowNewQueues = case lookupValue "TRANSPORT" "new_queues" ini of
+                      Right "on" -> True
+                      Right "off" -> False
+                      Right _ -> error "invalid new_queues setting in INI file"
+                      _ -> True,
+                    newQueueBasicAuth = case lookupValue "TRANSPORT" "create_password" ini of
+                      Right auth -> either error Just . strDecode $ encodeUtf8 auth
```
jr-simplex, line 111: make error with message?

tests/AgentTests/FunctionalAPITests.hs
```diff
@@ -127,6 +130,50 @@ functionalAPITests t = do
       testServerMatrix2 t testSwitchAsync
     describe "should delete connection during rotation" $
       testServerMatrix2 t testSwitchDelete
+  describe "SMP basic auth" $ do
+    describe "with server auth" $ do
+      --                                       allow NEW | server auth, v | clnt1 auth, v  | clnt2 auth, v    |  2 - success, 1 - JOIN fail, 0 - NEW fail
+      it "success                " $ testBasicAuth t True (Just "abcd", 5) (Just "abcd", 5) (Just "abcd", 5) `shouldReturn` 2
+      it "disabled               " $ testBasicAuth t False (Just "abcd", 5) (Just "abcd", 5) (Just "abcd", 5) `shouldReturn` 0
+      it "NEW fail, no auth      " $ testBasicAuth t True (Just "abcd", 5) (Nothing, 5) (Just "abcd", 5) `shouldReturn` 0
+      it "NEW fail, no auth both " $ testBasicAuth t True (Just "abcd", 5) (Nothing, 5) (Nothing, 5) `shouldReturn` 0
```
jr-simplex, line 139: not sure what's the point in distinguishing these test cases when `testBasicAuth` + `testCreateQueueAuth` are written in a way that makes the first impossible to succeed..

jr-simplex (approved):

apps/smp-server/Main.hs
epoberezkin: parser will supply the error

PR #565, opened 2022-11-17
refactor server CLIs

epoberezkin:
Other than the comment below, there should be no functional differences.

### DIFF
--- a/src/Simplex/Messaging/Server/CLI.hs
+++ b/src/Simplex/Messaging/Server/CLI.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
-{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeApplications #-}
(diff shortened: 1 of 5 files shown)


src/Simplex/Messaging/Server/CLI.hs
```diff
@@ -245,49 +167,46 @@ strictIni section key ini =
...
+    _ -> "Store log disabled."
+  forM_ transports $ \(p, ATransport t) ->
+    putStrLn $ "Listening on port " <> p <> " (" <> transportName t <> ")..."
 
-clearDirs :: ServerCLIConfig cfg -> IO ()
-clearDirs ServerCLIConfig {cfgDir, logDir} = do
-  clearDirIfExists cfgDir
-  clearDirIfExists logDir
-  where
-    clearDirIfExists path = whenM (doesDirectoryExist path) $ listDirectory path >>= mapM_ (removePathForcibly . combine path)
```
epoberezkin, line 288: I didn't quite understand if there was a reason to keep the actual folders, rather than just delete them on initialisation, this difference is not preserved.

src/Simplex/Messaging/Server/CLI.hs
```diff
@@ -76,152 +40,110 @@ confirmOrExit s = do
...
+        \CN = SMP server CA\n\
+        \O = SimpleX\n\n\
+        \[v3]\n\
+        \subjectKeyIdentifier = hash\n\
+        \authorityKeyIdentifier = keyid:always\n\
+        \basicConstraints = critical,CA:true\n"
+    -- TODO revise https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.3, https://www.rfc-editor.org/rfc/rfc3279#section-2.3.5
+    -- IP and FQDN can't both be used as server address interchangeably even if IP is added
+    -- as Subject Alternative Name, unless the following validation hook is disabled:
+    -- https://hackage.haskell.org/package/x509-validation-1.6.10/docs/src/Data-X509-Validation.html#validateCertificateName
```
jr-simplex, line 117:
```suggestion
```

src/Simplex/Messaging/Server/CLI.hs
jr-simplex: I think it's ok

jr-simplex (approved):

PR #566, opened 2022-11-18
initialize SMP server interactively, CLI options

epoberezkin:


src/Simplex/Messaging/Server/CLI.hs
```diff
@@ -33,12 +35,10 @@ exitError :: String -> IO ()
...
-confirmOrExit s = do
-  putStrLn s
-  putStr "Continue (Y/n): "
-  hFlush stdout
-  ok <- getLine
-  when (ok /= "Y") exitFailure
+confirmOrExit s =
+  withPrompt (s <> "\nContinue (Y/n): ") $ do
+    ok <- getLine
+    when (ok /= "Y") exitFailure
```
jr-simplex, line 41: it's better to print something, e.g. "Aborting..." before `exitFailure` - it's not entirely clear if server files weren't deleted if you type "y" lowercase for example.. (it just does nothing). Or at least compare with it as well
```suggestion
    when (ok /= "Y" && ok /= "y") exitError "Aborting deletion"
```

jr-simplex, line 41: It's better to print something before `exitFailure`, it's not entirely clear files aren't deleted if you type "y" lowercase.. (it just does nothing). Or at least compare with it as well.
"Aborting deletion" can be passed to `msg` here.
```suggestion
confirmOrExit :: String -> String -> IO ()
confirmOrExit s msg =
  withPrompt (s <> "\nContinue (Y/n): ") $ do
    ok <- getLine
    when (ok /= "Y" && ok /= "y") $ exitError msg
```

src/Simplex/Messaging/Server/CLI.hs
```diff
@@ -174,6 +174,24 @@ iniOnOff section name ini = case lookupValue section name ini of
   Right s -> error . T.unpack $ "invalid INI setting " <> name <> ": " <> s
   _ -> Nothing
 
+withPrompt :: String -> IO a -> IO a
+withPrompt s a = putStr s >> hFlush stdout >> a
+
+onOffPrompt :: String -> Bool -> IO Bool
+onOffPrompt prompt def =
+  withPrompt (prompt <> if def then " (Yn): " else " (yN): ") $
```
jr-simplex, line 182:
```suggestion
  withPrompt (prompt <> if def then " (Y/n): " else " (y/N): ") $
```

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -44,62 +56,99 @@ smpServerCLI cfgPath logPath =
...
-      let x509cfg = defaultX509Config {commonName = fromMaybe ip fqdn, signAlgorithm}
-      fp <- createServerX509 cfgPath x509cfg
-      writeFile iniFile iniFileContent
-      putStrLn $ "Server initialized, you can modify configuration in " <> iniFile <> ".\nRun `" <> executableName <> " start` to start server."
-      printServiceInfo serverVersion fp
-      warnCAPrivateKeyFile cfgPath x509cfg
+    initializeServer opts
+      | scripted opts = initialize opts
+      | otherwise = do
+        void $ withPrompt "SMP server will be initialized (press Enter)" getLine
```
jr-simplex, line 62: anticipating questions
```suggestion
      | otherwise = do
        putStrLn "You can run this non interactively using -y option"
        void $ withPrompt "SMP server will be initialized (press Enter)" getLine
```

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -44,62 +56,99 @@ smpServerCLI cfgPath logPath =
...
+        enableStoreLog <- onOffPrompt "Enable store log to restore queues and messages on server restart" True
+        logStats <- onOffPrompt "Enable server storing server statistics daily" False
+        putStrLn "Should server require a password to create new messaging queues?"
+        password <- withPrompt "Enter 'r' for random (default), 'n' - no password, or enter password: " serverPassword
+        let host = fromMaybe (ip opts) (fqdn opts)
+        host' <- withPrompt ("Enter server FQDN or IP address for certificate (" <> host <> "): ") getLine
```
jr-simplex, line 68:
```suggestion
        host' <- withPrompt ("Enter server FQDN or IP address for certificate (default " <> host <> "): ") getLine
```

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -182,18 +244,39 @@ cliCommandP cfgPath logPath iniFile =
...
+              ( long "password"
+                  <> metavar "PASSWORD"
+                  <> help "Set password to create new messaging queues"
+                  <> value SPRandom
+              )
+      scripted <-
+        switch
+          ( long "yes"
+              <> short 'y'
+              <> help "Scripted initialization using command-line options"
```
jr-simplex, line 278: just not used to this language, maybe it's ok
```suggestion
              <> help "Non interactive initialization using command-line options"
```

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -44,62 +56,99 @@ smpServerCLI cfgPath logPath =
...
-      createDirectoryIfMissing True logPath
-      let x509cfg = defaultX509Config {commonName = fromMaybe ip fqdn, signAlgorithm}
-      fp <- createServerX509 cfgPath x509cfg
-      writeFile iniFile iniFileContent
-      putStrLn $ "Server initialized, you can modify configuration in " <> iniFile <> ".\nRun `" <> executableName <> " start` to start server."
-      printServiceInfo serverVersion fp
-      warnCAPrivateKeyFile cfgPath x509cfg
+    initializeServer opts
+      | scripted opts = initialize opts
+      | otherwise = do
```
jr-simplex, line 61: For clarity there could be lines after each question:
"You selected: yes / no / random / etc."
 - just to be sure that you selected what was intended, e.g. after typing "y" instead of offered "Y".

Or/and print summary in the end. Not necessary though

jr-simplex (dismissed):

src/Simplex/Messaging/Server/CLI.hs
epoberezkin: It's not changed really, and we didn't accept lowercase Y before - just in case

src/Simplex/Messaging/Server/CLI.hs
epoberezkin: it's not exactly the same as above - although might be confusing. Here, there is the default, in the above case the input is required.

src/Simplex/Messaging/Server/Main.hs
epoberezkin: Don't think we need to print it, it is available in help

src/Simplex/Messaging/Server/Main.hs
epoberezkin: it's only going to be if passed via options, or it'll be 127.0.0.1

src/Simplex/Messaging/Server/Main.hs
epoberezkin: added -h option info

src/Simplex/Messaging/Server/CLI.hs
epoberezkin: added output that nothing was deleted

PR #577, opened 2022-12-08
agent: api to get hash of ratchet associated data for connection verification

epoberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -248,6 +249,10 @@ deleteConnection c = withAgentEnv c . deleteConnection' c
 getConnectionServers :: AgentErrorMonad m => AgentClient -> ConnId -> m ConnectionStats
 getConnectionServers c = withAgentEnv c . getConnectionServers' c
 
+-- | get connection ratchet associated data for connection verification (should match AD of the peer)
```
jr-simplex, line 252:
```suggestion
-- | get hash of connection ratchet associated data for connection verification (should match AD of the peer)
```

PR #581, opened 2022-12-26
option to enable/disable TLS handshake error logs (disable by default)

epoberezkin:


src/Simplex/Messaging/Notifications/Server/Main.hs
```diff
@@ -71,12 +71,13 @@ ntfServerCLI cfgPath logPath =
           \# and restoring it when the server is started.\n\
           \# Log is compacted on start (deleted objects are removed).\n"
             <> ("enable: " <> onOff enableStoreLog <> "\n\n")
-            <> "log_stats: off\n\n"
-            <> "[TRANSPORT]\n"
-            <> "# host is only used to print server address on start\n"
+            <> "log_stats: off\n\n\
+               \[TRANSPORT]\n\
+               \# host is only used to print server address on start\n"
```
jr-simplex, line 76: this seems more error prone - in addition to \n you have to not forget slashes both in beginning and end, but only on specific lines. I don't think it's a good use for multiline string, concatenation seems more straightforward here.

why change this at all if it's working, if it's not tested miss a symbol and it may fail for new installations

src/Simplex/Messaging/Notifications/Server/Main.hs
```diff
@@ -71,12 +71,13 @@ ntfServerCLI cfgPath logPath =
...
             <> ("host: " <> host <> "\n")
             <> ("port: " <> defaultServerPort <> "\n")
-            <> "websockets: off\n"
+            <> "log_tls_errors: off\n\
+               \websockets: off\n"
```
jr-simplex, line 80: just do normal concatenation here as well?

```suggestion
            <> "log_tls_errors: off\n
            <> "websockets: off\n"
```

jr-simplex, line 80: just do normal concatenation here as well?

```suggestion
            <> "log_tls_errors: off\n
            <> "websockets: off\n"
```

src/Simplex/Messaging/Notifications/Server/Main.hs
jr-simplex: we may have tests for this now though as of recently, still

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -109,26 +109,27 @@ smpServerCLI cfgPath logPath =
...
-                <> "[AUTH]\n"
-                <> "# Set new_queues option to off to completely prohibit creating new messaging queues.\n"
-                <> "# This can be useful when you want to decommission the server, but not all connections are switched yet.\n"
-                <> "new_queues: on\n\n"
-                <> "# Use create_password option to enable basic auth to create new messaging queues.\n"
-                <> "# The password should be used as part of server address in client configuration:\n"
-                <> "# smp://fingerprint:password@host1,host2\n"
-                <> "# The password will not be shared with the connecting contacts, you must share it only\n"
-                <> "# with the users who you want to allow creating messaging queues on your server.\n"
+                <> "[AUTH]\n\
```
jr-simplex, line 112: why change what's working

src/Simplex/Messaging/Notifications/Server/Main.hs
epoberezkin: I don't see how concatenation is any less error-prone, seems the opposite. In one case you need to type `", eol, <,>, "`, in another: `\, eol, \` - I am not sure how 5 characters are less error-prone, in both cases it either compiles or it doesn't, in either case if you miss `\n` it will not break the line in the output.

src/Simplex/Messaging/Server/Main.hs
epoberezkin: what happened with the argument for consistency. it's not even refactoring, just plain replacement of many concatenated strings with one.

PR #585, opened 2022-12-27
send/process "quota exceeded" message from SMP server when sender gets ERR QUOTA

epoberezkin:


src/Simplex/Messaging/Protocol.hs
```diff
@@ -328,44 +334,72 @@ instance StrEncoding RcvMessage where
...
+  | RcvMsgQuota
+      { msgTs :: SystemTime
+      }
 
-encodeRcvMsgBody :: RcvMsgBody -> C.MaxLenBS MaxRcvMessageLen
-encodeRcvMsgBody RcvMsgBody {msgTs, msgFlags, msgBody} =
-  let rcvMeta :: C.MaxLenBS 16 = C.unsafeMaxLenBS $ smpEncode (msgTs, msgFlags, ' ')
-   in C.appendMaxLenBS rcvMeta msgBody
+msgQuotaTag :: ByteString
+msgQuotaTag = "QUOTA "
```
jr-simplex, line 348: not sure about including space

PR #589, opened 2022-12-30
FileStore

avently:


src/Simplex/FileTransfer/Server/Store.hs
```diff
@@ -1 +1,99 @@
...
+data FileStore = FileStore
+  { chunks :: TMap SenderId FileRec,
+    recipients :: TMap RecipientId (SenderId, RcvPublicVerifyKey)
+  }
+
+data FileRec = FileRec
+  { senderId :: SenderId,
+    senderKey :: SndPublicVerifyKey,
+    readers :: TVar [RecipientId],
+    filepath :: String
```
epoberezkin, line 38:
```suggestion
    filePath :: TVar (Maybe FilePath)
```

It's created when there is no file yet, file is added with another command. FilePath is a type synonym for String.

src/Simplex/FileTransfer/Server/Store.hs
```diff
@@ -1 +1,99 @@
...
+
+data FileStore = FileStore
+  { chunks :: TMap SenderId FileRec,
+    recipients :: TMap RecipientId (SenderId, RcvPublicVerifyKey)
+  }
+
+data FileRec = FileRec
+  { senderId :: SenderId,
+    senderKey :: SndPublicVerifyKey,
+    readers :: TVar [RecipientId],
```
epoberezkin, line 37:
```suggestion
    rcvIds :: TVar (Set RecipientId),
```

readers is a bit confusing.

src/Simplex/FileTransfer/Server/Store.hs
epoberezkin: which means there will be one more function to add this path.

PR #615, opened 2023-01-18
end SMP client connection after configured number of PING errors, reset PING error count on any success

epoberezkin:


jr-simplex:
"end SMP client connection after configured number of PING errors" - this doesn't seem to happen, only that ping stops

src/Simplex/Messaging/Client.hs
```diff
@@ -308,9 +314,20 @@ getProtocolClient protocolServer cfg@ProtocolClientConfig {qSize, networkConfig,
...
+      r_ <- runExceptT (sendProtocolCommand c Nothing "" protocolPing)
+      if smpPingCount networkConfig == 0
+        then ping c
+        else case r_ of
+          Right _ -> resetLoop
+          -- below is to avoid resetting connection on incorrect Ntf server responses, it can be removed after the server is updated
+          Left (PCEProtocolError AUTH) -> resetLoop
+          Left _ -> do
+            cnt <- atomically $ stateTVar pingErrorCount $ \cnt -> (cnt + 1, cnt + 1)
+            when (cnt < smpPingCount networkConfig) $ ping c
```
jr-simplex, line 328: should the opposite check be made elsewhere to end SMP client connection?

PR #646, opened 2023-02-20
xftp: add server to client encryption layer with cryptobox

epoberezkin:


src/Simplex/FileTransfer/Transport.hs
```diff
@@ -40,18 +46,33 @@ sendFile h send = go
...
+    go sbState 0 = do
+      -- TODO remove padding when HTTP2 issue is fixed
+      let authTag = BA.convert (LC.sbAuth sbState) <> B.replicate (xftpBlockSize - C.authTagSize) '#'
+      send $ byteString authTag
+    go sbState sz =
+      getFileChunk h sz >>= \ch -> do
+        let (encCh, sbState') = LC.sbEncryptChunk sbState ch
+        send (byteString encCh) `E.catch` \(e :: E.SomeException) -> print e >> E.throwIO e
+        -- TODO remove delay when HTTP2 issue is fixed
+        threadDelay 500
```
spaced4ndy, line 65: what's the HTTP2 issue that requires padding and delay?

epoberezkin: https://github.com/kazu-yamamoto/http2/issues/40

PR #647, opened 2023-02-22
xftp: sender file description

spaced4ndy:


tests/XFTPCLI.hs
```diff
@@ -60,16 +71,24 @@ testXFTPCLISendReceive2servers = withXFTPServer . withXFTPServer2 $ do
   xftp ["rand", filePath, "19mb"] `shouldReturn` ["File created: " <> filePath]
   file <- LB.readFile filePath
   getFileSize filePath `shouldReturn` 19 * mb
-  let fd1 = filePath <> ".xftp" </> "rcv1.xftp"
-      fd2 = filePath <> ".xftp" </> "rcv2.xftp"
+  let fdRcv1 = filePath <> ".xftp" </> "rcv1.xftp"
+      fdRcv2 = filePath <> ".xftp" </> "rcv2.xftp"
+      fdSnd = filePath <> ".xftp" </> "snd.xftp"
```
epoberezkin, line 76: I don't like the idea of putting it in the same folder, it's a bit risky, also I was going to make that in case it's one recipient it will be a file, not a folder.

I suggest we name the file `filePath <> ".xftp.private"` or `filePath <> ".xftp.sender"` (with uniqueCombine obv.)

src/Simplex/FileTransfer/Client/Main.hs
```diff
@@ -316,38 +323,61 @@ cliSendFile SendOptions {filePath, outputDir, numRecipients, xftpServers, retryC
...
+    writeFileDescriptions :: String -> [FileDescription 'FPRecipient] -> FileDescription 'FPSender -> IO ([FilePath], FilePath)
+    writeFileDescriptions fileName fdRcvs fdSnd = do
       outDir <- uniqueCombine (fromMaybe "." outputDir) (fileName <> ".xftp")
       createDirectoryIfMissing True outDir
-      forM (zip [1 ..] fds) $ \(i :: Int, fd) -> do
+      fdRcvPaths <- forM (zip [1 ..] fdRcvs) $ \(i :: Int, fd) -> do
         let fdPath = outDir </> ("rcv" <> show i <> ".xftp")
         B.writeFile fdPath $ strEncode fd
         pure fdPath
+      let fdSndPath = outDir </> "snd.xftp"
```
epoberezkin, line 350:
```suggestion
      let fdSndPath = outDir </> "snd.xftp.private"
```

PR #648, opened 2023-02-22
xftp: FDEL and FACK commands

epoberezkin:


tests/XFTPServerTests.hs
```diff
@@ -72,3 +76,53 @@ runTestFileChunkDelivery s r = do
...
+  (sId, [rId]) <- createXFTPChunk s spKey file [rcvKey]
+  uploadXFTPChunk s spKey sId chunkSpec
+
+  downloadXFTPChunk r rpKey rId $ XFTPRcvChunkSpec "tests/tmp/received_chunk1" chSize digest
+  liftIO $ B.readFile "tests/tmp/received_chunk1" `shouldReturn` bytes
+  ackXFTPChunk r rpKey rId
+  downloadXFTPChunk r rpKey rId (XFTPRcvChunkSpec "tests/tmp/received_chunk2" chSize digest)
+    `catchError` (liftIO . (`shouldBe` PCEProtocolError AUTH))
+  ackXFTPChunk r rpKey rId
+    `catchError` (liftIO . (`shouldBe` PCEProtocolError AUTH))
```
spaced4ndy, line 128: what happens when all recipients acknowledge the chunk? should server delete the file?

epoberezkin: no, if we are going to allow downloading with sender ID

PR #649, opened 2023-02-22
xftp: cli - delete file command, acknowledge received chunks

spaced4ndy:


tests/XFTPCLI.hs
```diff
@@ -99,4 +102,37 @@ testXFTPCLISendReceive2servers = withXFTPServer . withXFTPServer2 $ do
...
+                     fdRcv2,
+                     "Sender file description:",
+                     fdSnd
+                   ]
+  xftp ["del", fdRcv1]
+    `shouldThrow` anyException
+  xftp ["recv", fdRcv1, recipientFiles, "--tmp=tests/tmp"]
+    `shouldReturn` ["File received: " <> recipientFiles </> "testfile"]
+  LB.readFile (recipientFiles </> "testfile") `shouldReturn` file
+  listDirectory xftpServerFiles >>= \fs -> fs `shouldNotBe` []
```
epoberezkin, line 129: You can check file count instead

tests/XFTPCLI.hs
```diff
@@ -99,4 +102,37 @@ testXFTPCLISendReceive2servers = withXFTPServer . withXFTPServer2 $ do
...
+  listDirectory xftpServerFiles2 >>= \fs -> fs `shouldNotBe` []
+  xftp ["del", fdSnd]
+    `shouldReturn` ["File deleted"]
+  listDirectory xftpServerFiles >>= \fs -> fs `shouldBe` []
```
epoberezkin, line 133:
```suggestion
  listDirectory xftpServerFiles >>= (`shouldBe` [])
```

tests/XFTPCLI.hs
```diff
@@ -55,6 +56,8 @@ testXFTPCLISendReceive = withXFTPServer $ do
   testInfoFile fdRcv2 "Recipient"
   testReceiveFile fdRcv2 "testfile_1" file
   testInfoFile fdSnd "Sender"
+  xftp ["recv", fdSnd, recipientFiles, "--tmp=tests/tmp"]
+    `shouldThrow` anyException
```
epoberezkin, line 60: It’s better to check for specific exception maybe?
Also - why would CLI have throw an exception at all - should it not print error?

tests/XFTPCLI.hs
```diff
@@ -99,4 +102,37 @@ testXFTPCLISendReceive2servers = withXFTPServer . withXFTPServer2 $ do
...
+    `shouldReturn` ["File received: " <> recipientFiles </> "testfile"]
+  LB.readFile (recipientFiles </> "testfile") `shouldReturn` file
+  listDirectory xftpServerFiles >>= \fs -> fs `shouldNotBe` []
+  listDirectory xftpServerFiles2 >>= \fs -> fs `shouldNotBe` []
+  xftp ["del", fdSnd]
+    `shouldReturn` ["File deleted"]
+  listDirectory xftpServerFiles >>= \fs -> fs `shouldBe` []
+  listDirectory xftpServerFiles2 >>= \fs -> fs `shouldBe` []
+  xftp ["recv", fdRcv2, recipientFiles, "--tmp=tests/tmp"]
+    `shouldThrow` anyException
```
epoberezkin, line 136: Same

tests/XFTPCLI.hs
```diff
@@ -99,4 +102,37 @@ testXFTPCLISendReceive2servers = withXFTPServer . withXFTPServer2 $ do
...
+  xftp ["send", filePath, senderFiles, "-n", "2", "-s", testXFTPServerStr <> ";" <> testXFTPServerStr2, "--tmp=tests/tmp"]
+    `shouldReturn` [ "File uploaded!",
+                     "Pass file descriptions to the recipient(s):",
+                     fdRcv1,
+                     fdRcv2,
+                     "Sender file description:",
+                     fdSnd
+                   ]
+  xftp ["del", fdRcv1]
+    `shouldThrow` anyException
```
epoberezkin, line 125: Better check specific one. Also - why would it throw?

tests/XFTPCLI.hs
spaced4ndy: I didn't find a more appropriate function

tests/XFTPCLI.hs
spaced4ndy: when run as an executable it prints a correct error, then exits with an error code. in tests trying to check for output fails with "uncaught exception: ExitCode", probably due to how capture_ / withArgs work

tests/XFTPCLI.hs
spaced4ndy: delete command expects sender file description

PR #651, opened 2023-02-23
xftp: expire files on the server, track/limit used storage

epoberezkin:


src/Simplex/FileTransfer/Server/Env.hs
```diff
@@ -29,6 +35,14 @@ data XFTPServerConfig = XFTPServerConfig
...
     storeLogFile :: Maybe FilePath,
     filesPath :: FilePath,
+    -- | set to False to prohibit creating new queues
+    fileSizeQuota :: Maybe Int64,
+    -- | simple password that the clients need to pass in handshake to be able to create new queues
+    allowNewFiles :: Bool,
+    -- | server storage quota
+    newFileBasicAuth :: Maybe BasicAuth,
+    -- | time after which the messages can be removed from the queues and check interval, seconds
+    fileExpiration :: Maybe ExpirationConfig,
```
spaced4ndy, line 45:
```suggestion
    -- | server storage quota
    fileSizeQuota :: Maybe Int64,
    -- | set to False to prohibit creating new files
    allowNewFiles :: Bool,
    -- | simple password that the clients need to pass in handshake to be able to create new files
    newFileBasicAuth :: Maybe BasicAuth,
    -- | time after which the files can be removed and check interval, seconds
    fileExpiration :: Maybe ExpirationConfig,
```

PR #661, opened 2023-02-27
xftp: server stats

epoberezkin:


src/Simplex/FileTransfer/Server.hs
```diff
@@ -248,6 +252,9 @@ processXFTPRequest HTTP2Body {bodyPart} = \case
         withFileLog $ \sl -> do
           logAddFile sl sId file ts
           logAddRecipients sl sId rcps
+        stats <- asks serverStats
+        atomically $ modifyTVar' (filesCreated stats) (+ 1)
+        atomically $ modifyTVar' (fileRecipients stats) (+ length rks)
```
spaced4ndy, line 257: why are these and other stats updates in separate atomically?

spaced4ndy (approved):

src/Simplex/FileTransfer/Server.hs
epoberezkin: same as in SMP server really, there is no much benefit in them being strongly consistent, but it reduces the deadlock risks.

PR #662, opened 2023-02-28
Correct minor Docker syntax error

sethforprivacy:
Noticed that the environment variables for SMP server when using Docker were improperly quoted, so I corrected them.

This will fix an issue with server address being displayed incorrectly in the CLI when viewing init logs.

I also explicitly added the `:latest` tag for the Docker image in the Running example to make it more clear which tag is being run (no tag defaults to `latest`.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,15 +113,15 @@ On Linux, you can deploy smp server using Docker. This will download image from 
    mkdir -p ~/simplex/{config,logs}
    ```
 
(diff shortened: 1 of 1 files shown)


shumvgolove:
LGTM, thanks for correcting the docs.

epoberezkin:
@sethforprivacy thank you!

sethforprivacy:
Thank you all for the excellent docs, made getting an SMP server up and running a breeze!

PR #665, opened 2023-03-01
xftp: remove file description that cannot be used

epoberezkin:


src/Simplex/FileTransfer/Client/Main.hs
```diff
@@ -588,6 +596,26 @@ withRetry retryCount = withRetry' retryCount . withExceptT (CLIError . show)
         logWarn ("retrying: " <> tshow e)
         withRetry' (n - 1) a
 
+removeFD :: Bool -> FilePath -> IO ()
+removeFD yes fd
+  | yes = do
+    putStrLn $ "\nFile description " <> fd <> " is deleted."
+    removeFile fd
```
spaced4ndy, line 603: why is print before action?
```suggestion
    removeFile fd
    putStrLn $ "\nFile description " <> fd <> " is deleted."
```

spaced4ndy (approved):

PR #674, opened 2023-03-07
xftp: agent send file design wip

spaced4ndy:


src/Simplex/FileTransfer/Agent.hs
```diff
@@ -179,3 +182,66 @@ runXFTPLocalWorker c@AgentClient {subQ} doWork = do
...
+      -- calculate digest
+      -- prepare chunk specs
+      -- db:
+      --   - update file status to Encrypted
+      --   - create chunks according to chunk specs
+      -- ? since which servers are online is unknown,
+      -- ? we can't blindly assign servers to replicas.
+      -- ? should we XFTP create chunks on servers here,
+      -- ? with retrying for different servers,
+      -- ? keeping a list of servers that were tried?
```
epoberezkin, line 221: yes, that's what we do when we create connection - similar logic can be used here

epoberezkin: I don't think it's necessary to keep tried servers in db though, it'll be just during the session

epoberezkin:
So my understanding is that we will request to upload a file on a given path, and once complete the agent will return sender's and recipients' file descriptions in one event, that chat will save to the database.

We probably need file ID straight away (same for receiving) so we can request the state / descriptions later - it's possible that the agent sends the events but chat fails to store them, so on chat start we should be "subscribing" on all incomplete files and agent returning the current state, including existing descriptions if necessary.

I will be adding these events to the agent types tomorrow.

src/Simplex/FileTransfer/Agent.hs
spaced4ndy: yes, the intention was to keep servers list in memory, could even be for chunk retries

PR #675, opened 2023-03-09
xftp: agent error handling

spaced4ndy:
- chunk download doesn't loop on permanent errors
- decryption errors are considered permanent - local worker doesn't retry
- update replica retries; to do - consider use for this field, or remove it
- rcv file Error status - to prevent repeat reads of chunks for download, files for decryption; also plan to use it for filtering on cleanup
- error string saved in separate field for debugging (not part of status type)
- agent event for rcv file errors

### DIFF
--- a/src/Simplex/FileTransfer/Agent.hs
+++ b/src/Simplex/FileTransfer/Agent.hs
@@ -13,10 +13,11 @@ module Simplex.FileTransfer.Agent
     receiveFile,
     addXFTPWorker,
     -- Sending files
-    sendFile,
+    _sendFile,
   )
 where
 
+import Control.Logger.Simple (logError)
 import Control.Monad
 import Control.Monad.Except
 import Control.Monad.Reader
@@ -31,7 +32,8 @@ import Simplex.FileTransfer.Types
 import Simplex.FileTransfer.Util (uniqueCombine)
 import Simplex.Messaging.Agent.Client
 import Simplex.Messaging.Agent.Env.SQLite
-import Simplex.Messaging.Agent.Protocol (ACommand (FRCVD), AParty (..), AgentErrorType (INTERNAL))
+import Simplex.Messaging.Agent.Protocol
+import qualified Simplex.Messaging.Agent.Protocol as AP
 import Simplex.Messaging.Agent.RetryInterval
 import Simplex.Messaging.Agent.Store
 import Simplex.Messaging.Agent.Store.SQLite
@@ -39,7 +41,7 @@ import qualified Simplex.Messaging.Crypto.Lazy as LC
 import Simplex.Messaging.Encoding
 import Simplex.Messaging.Protocol (XFTPServer)
 import qualified Simplex.Messaging.TMap as TM
-import Simplex.Messaging.Util (whenM)
+import Simplex.Messaging.Util (tshow, whenM)
 import UnliftIO
 import UnliftIO.Directory
 import qualified UnliftIO.Exception as E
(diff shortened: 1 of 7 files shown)


src/Simplex/FileTransfer/Agent.hs
```diff
@@ -77,31 +79,45 @@ runXFTPWorker c srv doWork = do
...
-    downloadFileChunk RcvFileChunk {userId, rcvFileId, rcvChunkId, chunkNo, chunkSize, digest, fileTmpPath, replicas = replica : _} = do
+            downloadFileChunk fc replica
+              `catchError` retryOnError delay' loop (workerInternalError c rcvFileId . show)
+          where
+            retryOnError :: Int -> m () -> (AgentErrorType -> m ()) -> AgentErrorType -> m ()
+            retryOnError chunkDelay loop done e = do
+              logError $ "XFTP worker error: " <> tshow e
+              case e of
+                BROKER _ NETWORK -> retryLoop
+                BROKER _ TIMEOUT -> retryLoop
```
epoberezkin, line 101: why not use the existing function for agent errors?

epoberezkin: possibly with some additional conditions.

src/Simplex/FileTransfer/Agent.hs
```diff
@@ -113,7 +129,14 @@ runXFTPWorker c srv doWork = do
...
           all (\RcvFileChunk {replicas} -> any received replicas) chunks
-    downloadFileChunk _ = throwError $ INTERNAL "no replica"
+
+workerInternalError :: AgentMonad m => AgentClient -> RcvFileId -> String -> m ()
+workerInternalError c rcvFileId internalErrStr = do
+  withStore' c $ \db -> updateRcvFileError db rcvFileId internalErrStr
+  notifyInternalError c rcvFileId internalErrStr
+
+notifyInternalError :: (MonadUnliftIO m) => AgentClient -> RcvFileId -> String -> m ()
+notifyInternalError AgentClient {subQ} rcvFileId internalErrStr = atomically $ writeTBQueue subQ ("", "", AP.FRCVERR rcvFileId $ AP.INTERNAL internalErrStr)
```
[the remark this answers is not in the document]
epoberezkin: actually, the whole type should be already imported here.

src/Simplex/FileTransfer/Agent.hs
```diff
@@ -31,15 +32,16 @@ import Simplex.FileTransfer.Types
 import Simplex.FileTransfer.Util (uniqueCombine)
 import Simplex.Messaging.Agent.Client
 import Simplex.Messaging.Agent.Env.SQLite
-import Simplex.Messaging.Agent.Protocol (ACommand (FRCVD), AParty (..), AgentErrorType (INTERNAL))
+import Simplex.Messaging.Agent.Protocol
+import qualified Simplex.Messaging.Agent.Protocol as AP
```
epoberezkin, line 36: I am not in support of namespacing all imports

epoberezkin: also, this both imports all and also gives the name

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -328,6 +329,7 @@ data ACommandTag (p :: AParty) where
   ERR_ :: ACommandTag Agent
   SUSPENDED_ :: ACommandTag Agent
   FRCVD_ :: ACommandTag Agent
+  FRCVERR_ :: ACommandTag Agent
```
epoberezkin, line 332: I think it should be just inside ERR with subtype

spaced4ndy: discussed, to do - decide on where to include file id in event, possibly generalize conn id to entity id

PR #677, opened 2023-03-09
xftp: cleanup complete & error files tmp paths; remove save path before decrypting if exists

spaced4ndy:


src/Simplex/FileTransfer/Agent.hs
```diff
@@ -151,8 +151,12 @@ runXFTPLocalWorker c@AgentClient {subQ} doWork = do
...
     noWorkToDo = void . atomically $ tryTakeTMVar doWork
     decryptFile :: RcvFile -> m ()
-    decryptFile RcvFile {rcvFileId, key, nonce, tmpPath, saveDir, chunks} = do
-      -- TODO remove tmpPath if exists
+    decryptFile RcvFile {rcvFileId, key, nonce, tmpPath, saveDir, savePath, chunks} = do
+      -- currently this check is obsolete, because getNextRcvFileToDecrypt filters by RFSReceived status
+      -- and therefore file can't have savePath here - maybe better to remove this logic altogether
+      forM_ savePath $ \p -> do
+        removePath p
+        withStore' c (`updateRcvFileNoSavePath` rcvFileId)
```
spaced4ndy, line 159: unsure about keeping or removing this logic - currently it's unreachable, but can become useful if for example store method filter changes

PR #687, opened 2023-03-15
xftp: streaming file encryption/decryption to avoid memory spikes

epoberezkin:


src/Simplex/Messaging/Crypto/Lazy.hs
```diff
@@ -158,16 +173,26 @@ sbInit_ secret nonce = (state2,) <$> cryptoPassed (Poly1305.initialize rs)
     state1 = XSalsa.derive state0 iv1
     (rs :: ByteString, state2) = XSalsa.generate state1 32
 
+sbEncryptChunkLazy :: SbState -> LazyByteString -> (LazyByteString, SbState)
```
spaced4ndy, line 176: this function seems to be unused

spaced4ndy (approved):

PR #704, opened 2023-04-02
xftp: agent API to set and test servers

epoberezkin:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -714,13 +729,60 @@ runSMPServerTest c userId (ProtoServerWithAuth srv auth) = do
...
+        incClientStat c userId xftp "XFTP_TEST" "OK"
+        pure $ either Just (const Nothing) r <|> maybe (Just (ProtocolTestFailure TSDisconnect $ BROKER addr TIMEOUT)) (const Nothing) ok
+      Left e -> pure (Just $ testErr TSConnect e)
+  where
+    addr = B.unpack $ strEncode srv
+    testErr :: ProtocolTestStep -> XFTPClientError -> ProtocolTestFailure
+    testErr step = ProtocolTestFailure step . protocolClientError XFTP addr
+    chSize :: Integral a => a
+    chSize = kb 256
+    getTempFilePath :: FilePath -> m FilePath
```
spaced4ndy, line 774: maybe make a utility function of it to re-use in xftp receive, and later send?

spaced4ndy: or maybe better to do it later, when send is done

spaced4ndy (approved):

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: I looked at it, and it uses different extension, yes - can be refactored out later.

PR #705, opened 2023-04-04
ci: disable 2 tests on linux only, switch to ubuntu 20 and 22

epoberezkin:


.github/workflows/build.yml
```diff
@@ -80,10 +81,10 @@ jobs:
...
           files: |
             LICENSE
-            smp-server-ubuntu-20_04-x86-64
-            ntf-server-ubuntu-20_04-x86-64
-            xftp-server-ubuntu-20_04-x86-64
-            xftp-ubuntu-20_04-x86-64
+            smp-server-ubuntu--$platform_name
+            ntf-server-ubuntu--$platform_name
+            xftp-server-ubuntu--$platform_name
+            xftp-ubuntu--$platform_name
```
spaced4ndy, line 87:
```suggestion
            smp-server-ubuntu-$platform_name
            ntf-server-ubuntu-$platform_name
            xftp-server-ubuntu-$platform_name
            xftp-ubuntu-$platform_name
```

.github/workflows/build.yml
```diff
@@ -80,10 +81,10 @@ jobs:
           prerelease: true
```
spaced4ndy, line 81: there's an if on matrix.os on line 73 that will throw away binaries for 22.04 - why prepare them? - as test?

PR #734, opened 2023-04-23
Add GH Actions workflow to build and push Docker image

### LINKED ISSUE #732: Image on Docker Hub not atomatically updated
The Docker image that's available on Docker Hub is currently on version 4.3.1, i.e. 3 versions behind. Do you guys have plans to keep it up to date, or is the intention that SMP server operators should build the Docker image themselves?

If it's of any help, I can make a PR with a GitHub Actions workflow to automatically build and push a Docker image on every release. Let me know what you think.

samsapti:
This PR adds a GitHub Actions workflow that will automatically build and push a Docker image to Docker Hub on every release. It creates four image tags:

- `vX`: The latest minor version under the specified major version
- `vX.Y`: The latest patch version under the specified minor version
- `vX.Y.Z`: The specified full version
- `latest`: Follows the latest release

It builds the image for Linux platform x64.

The only thing that you need to do is to:

- Set the repo secret `DOCKERHUB_USERNAME` to the username of your Docker Hub account
- Set the repo secret `DOCKERHUB_PASSWORD` to the password of your Docker Hub account

I have used a matrix to specify the application to be built and the Dockerfile that should be used. If you want to extend the workflow to build more applications, such as XFTP, just add a new entry to the matrix. Currently it only builds SimpleXMQ.

Closes #732.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -25,14 +25,6 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_PASSWORD }}
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
-        with:
-          platforms: linux/arm64
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
-
       - name: Extract metadata for Docker image
         id: meta
         uses: docker/metadata-action@v4
@@ -50,6 +42,5 @@ jobs:
         with:
           push: true
           file: ${{ matrix.dockerfile }}
-          platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,4 +1,4 @@
-name: Build and push SimpleXMQ Docker image
+name: Build and push Docker image to Docker Hub
 
 on:
   push:
@@ -7,8 +7,14 @@ on:
 
 jobs:
   build-and-push:
-    name: Build and push Docker image to Docker Hub
+    name: Build and push Docker image
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - application: simplexmq
+            dockerfile: build.Dockerfile
     steps:
       - name: Clone project
         uses: actions/checkout@v3
@@ -31,7 +37,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: ${{ secrets.DOCKERHUB_USERNAME }}/simplexmq
+          images: ${{ secrets.DOCKERHUB_USERNAME }}/${{ matrix.application }}
           flavor: |
             latest=auto
           tags: |
@@ -43,7 +49,7 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           push: true
-          file: build.Dockerfile
+          file: ${{ matrix.dockerfile }}
           platforms: linux/amd64,linux/arm64,linux/386
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v2
         with:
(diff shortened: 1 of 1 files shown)


shumvgolove:
Looks great, thanks for the quick update! Will merge this, once I'll test this setup a couple of times.

shumvgolove (commented):
Hey, I'm back with an update. Tested this setup with a [new repository ](https://hub.docker.com/repository/docker/simplexchat/smp-server) (unfortunately, existing repositories cannot be renamed in Docker hub, so I've recreated it) and while this PR is totally functional, this workflow will fail without some adjustments to `build.Dockerfile` - https://github.com/simplex-chat/simplexmq/pull/735/commits/9101fd550a5130a9892ab8b1501153eeb1ffeaea . Once this https://github.com/simplex-chat/simplexmq/pull/735 PR will be merged, we can merge this one too!

Oh, and a couple of things:
- Could you set target branch from `stable` to `master`? We will merge all this changes to `stable` afterwards, since this is our workflow :^)
- Please see small correction ↓

samsapti:
@shumvgolove I've changed the base branch and rebased this PR onto `master`, and I've renamed the application to `smp-server`.

> this workflow will fail without some adjustments to build.Dockerfile - https://github.com/simplex-chat/simplexmq/commit/9101fd550a5130a9892ab8b1501153eeb1ffeaea

I don't think that simply adding `--platform=$BUILDPLATFORM` will work. That will cause the image to always build for `amd64`/x86_64 (assuming the build is done from an `amd64` machine), unless you make some other changes to `build.Dockerfile` to actually cross-compile to `$TARGETPLATFORM`. Also, if you do that, then the `final` stage should use `--platform=$TARGETPLATFORM` (which is actually the default, so there's no need to specify that), because that's the stage that the image will use to run the container, so if it's always `amd64` it'll fail on other architectures.

Unless the build system you're using supports cross-compiling, you're better of just letting it use `--platform=$TARGETPLATFORM` (the default) for the `build` stage.

You might wanna take a look at [this](https://docs.docker.com/build/building/multi-platform).

samsapti:
Aha, that's because the `ubuntu` image is not available for the `linux/386` platform it seems. I included it in the workflow because the last release changelog stated added support for 32-bit platforms, but I forgot to check whether the image supports it. I'll remove it from the workflow.

shumvgolove:
Oh well, should’ve thought of that. Let me revert Docker changes first…

samsapti:
(rebased onto master)

samsapti:
(rebased again)

shumvgolove:
You will need to rebase one more time, sorry about that, currently fighting with `cabal` :^)

shumvgolove:
After multiple CI reruns, I have to say that, unfortunately, we need to disable arm64 as well. Not sure if it's insufficient resources for `qemu` emulation or networking issues on GitHub end, but CI is consistently [getting halted](https://github.com/shumvgolove/simplexmq/actions/runs/4790977478) while building `arm64` container with a somewhat obscure message: `The hosted runner: GitHub Actions 2 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.`

I'll give it a go one last time ([you can track the job here](https://github.com/shumvgolove/simplexmq/actions/runs/4794295439/jobs/8531081155)) just in case.

samsapti:
CircleCI uses a different format for workflow/pipeline files, so you would need another workflow. However, with CircleCI, you could define three CI jobs:

1) Build the x86_64 image on a x86_64 runner and push it as `simplexchat/smp-server:v<version>-amd64`
2) Build the ARMv8 image on a ARMv8 runner and push it as `simplexchat/smp-server:v<version>-arm64v8`
3) Create a Docker manifest to put both images into one and push it as `simplexchat/smp-server:v<version>`

I found an example of this [here](https://behroozam.medium.com/multi-arch-build-with-docker-buildx-and-circleci-900eb3827274).

> In this case, can we build `amd64` on Github free runners and arm64 on CircleCI simultaneously, or do we need to set up another Docker workflow just for the `arm64`?

I think it's easier to just do everything in CircleCI. It supports executing jobs in parallel, and you can specify that some jobs depend on other jobs (so from the example above, job 3 depends on 1 and 2, but 1 and 2 can run simultaneously because they don't depend on each other). If you build x86_64 on GitHub Actions and ARMv8 on CircleCI, you would have to do some fancy synchronization to have a third job somewhere that makes the Docker manifest only when both CI systems are done building the images.

shumvgolove:
Thanks for explanation. Let's proceed to merge this for the initial step and then figure out the best way to support `arm64` in the following issue: https://github.com/simplex-chat/simplexmq/issues/740

shumvgolove:
Yes, thanks for noticing!

PR #746, opened 2023-05-05
docker: rewrite entry point

tcely:
Rewritten to prefer understanding by inexperienced users:

- use single quotes wherever possible
- do not escape variable usage
- use explicit tests and logic
- always use braces for custom variables
- capitalize environment variables
- use consistent quoting

A password containing spaces should not cause errors.

When date fails, leave a file that explains what happened rather than a mystery.

Prefer `IP` over `ADDR` when both are provided.

This fixes how guessing if `addr` was an IP or not can go badly by letting the user explicitly specify that they want to use an IP address with a distinct environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,14 +113,15 @@ On Linux, you can deploy smp server using Docker. This will download image from 
    mkdir -p ~/simplex/{config,logs}
    ```
 
-2. Run your Docker container. You must change **your_ip_or_domain**. `-e "pass=password"` is optional variable to password-protect your `smp` server:
+2. Run your Docker container. You must provide **your_domain** using ADDR or **your_ip** using IP. `-e 'PASS=password'` is optional variable to password-protect your `smp` server:
    ```sh
    docker run -d \
-       -e "addr=your_ip_or_domain" \
-       -e "pass=password" \
+       -e 'ADDR=your_domain' \
+       -e 'IP=your_ip' \
+       -e 'PASS=password' \
        -p 5223:5223 \
-       -v $HOME/simplex/config:/etc/opt/simplex:z \
-       -v $HOME/simplex/logs:/var/opt/simplex:z \
+       -v "${HOME}/simplex/config:/etc/opt/simplex:z" \
+       -v "${HOME}/simplex/logs:/var/opt/simplex:z" \
        simplexchat/smp-server:latest
    ```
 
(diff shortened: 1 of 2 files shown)


tcely:
👋 to @shumvgolove as it looks like you made most of the docker change before.

Please review this, thanks.

shumvgolove:
Hey @tcely , thanks for your contribution, but, unfortunately, I'll NACK the "rewriting logic" part for several reasons.

`entrypoint` do not suppose to be read by inexperienced users, because it "hides" actual commands by this little abstraction (and it's used only in Docker, transparently). So, this commits:

- Adds unnecessary "if" statements, which just makes code more noisy for the sake of it. In the future, I'll update "addr" part a little bit, since we're planning to roll out `ipv6` support.
- Splits unified variable `addr` to separated instances, which adds complexity to documentation (minor) and to user (major). 
  > Prefer IP over ADDR when both are provided

  You can't do it in `smp-server`, therefore you can't do it in `entrypoint`, that's the point.

  > This fixes how guessing if addr was an IP or not can go badly

  Can you clarify how this can go badly (besides users misspelling their IP/FQDN)? Is the current logic incorrect (causes issues)? I've tested this thoroughly and didn't encounter any. Besides that, current behaviour mimics interactive initialization of `smp-server`:
  ```sh
  $ .smp-server init
  Use `smp-server init -h` for available options.
  SMP server will be initialized (press Enter)
  ...
  Enter server FQDN or IP address for certificate (127.0.0.1): 
  ```
- Makes redundant `date` check. I mean... we're talking about Docker here, specifically Ubuntu with coreutils installed, and, moreover, the command, which is [in POSIX specification](https://pubs.opengroup.org/onlinepubs/009695399/utilities/date.html), which **will** and **should** be in any Linux machine.
- Allows passwords that **can not** contain spaces: `Invalid password. Only latin letters, digits and symbols other than '@' and ':' are allowed`. In fact, `entrypoint` doesn't validate it either, we should add the check (I'll create the PR).
- Makes use of [Parameter Expansion](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02) inside double-quoted variables, which is unnecessary.

So yeah, I can't merge this in the current state, *but* I agree that we should {double,single}-quote and capitalize all external variables, which improves escaping and readability.

shumvgolove:
> I fully expect users to read the entrypoint script before using a docker image and to not use it, if they don't understand what the script does.

By the same extent, users should be reading `smp-server` code too, which I'm not expecting even from myself.

> users

`#define users`

System administrators/tinkerers? Maybe, if they want to contribute. End users? Nope. There's documentation how to quickly spin-up `smp-server` with Docker and that's what's important to them.

> I come down on the side of the reader, so I want it to be as clear as possible what is happening and not using anything that appears tricky

Not sure what's implied by "tricky". Is setting a parameter to `smp-server ` – tricky?

> Avoiding a "noisy" script is not a priority for me.

Well, it **is** for me, as I maintain it.

> 1) IDN that doesn't use the xn-- format

The domain suffix always use the english letters and the status quo catches this:
```sh
$ addr="äöü.com" pass="testing_testing" ./entrypoint
Certificate request self-signature ok
subject=CN = \C3\83\C2\A4\C3\83\C2\B6\C3\83\C2\BC.com
Server initialized, you can modify configuration in /etc/opt/simplex/smp-server.ini.
Run `smp-server start` to start server.
```

> 2) IPv6 contains letters

Yeah, It is an issue, but not currently. As I wrote:

> I'll update "addr" part a little bit, since we're planning to roll out ipv6 support.
 ---
>  As for date, there's no redundant check. It uses the output when it works, however it also provides useful information if it fails.

Could you provide such an example in the `Ubuntu 22.04 Docker` container? I'll be happy to fix it, if there's concrete results.

> You do want to avoid field splitting when passing arguments.

Field splitting only happens if variable is unquoted. Curly brackets needed in case if we're doing some expansion business, which is not happening here. I linked the POSIX specification in my first issue, in case you missed it. There **are** unquoted variables, and, as I've said:

>  I agree that we should {double,single}-quote and capitalize all external variables, which improves escaping and readability. 

Would you like to tackle this?

tcely:
> Would you like to tackle this?

Is there something I didn't already do in the entrypoint script for this?

shumvgolove:
Once again, NACK for the "rewriting logic" part, but ACK for "{double,single}-quote and capitalize all external variables". If you create new MR with only this changes (or force-push commits in this MR, doesn't matter), I'll be happy to merge this.

PR #748, opened 2023-05-05
DRY build.Dockerfile

tcely:
In addition to Don't Repeat Yourself, using ENV shows the versions used in the image history.


shumvgolove:
Thanks, that's good addition!

PR #749, opened 2023-05-06
configure message/file expiration time in INI file

epoberezkin:


src/Simplex/FileTransfer/Server/Main.hs
```diff
@@ -76,6 +77,8 @@ xftpServerCLI cfgPath logPath = do
           \# and restoring it when the server is started.\n\
           \# Log is compacted on start (deleted objects are removed).\n"
             <> ("enable: " <> onOff enableStoreLog <> "\n\n")
+            <> "# Expire messages that were not delivered for the specified number of days.\n"
```
spaced4ndy, line 80:
```suggestion
            <> "# Expire files after the specified number of days.\n"
```

PR #753, opened 2023-05-12
Docker entrypoint fixes

tcely:
Here are most of the changes from my previously NACK'd rewrite PR. #746 

~~Capitalized environment variables still needs to be done and the README will need updates for that part.~~

That work was done in #750 instead.

### DIFF
--- a/scripts/docker/entrypoint-smp-server
+++ b/scripts/docker/entrypoint-smp-server
@@ -29,7 +29,7 @@ fi
 # Backup store log just in case
 _file="${logd}/smp-server-store.log"
 if [ -f "${_file}" ]; then
-  _backup_extension="$(date --universal '+%FT%T%:z')"
+  _backup_extension="$(date --universal '+%FT%T')"
   cp -v -p "${_file}" "${_file}.${_backup_extension:-date-failed}"
   unset -v _backup_extension
 fi
--- a/scripts/docker/entrypoint-smp-server
+++ b/scripts/docker/entrypoint-smp-server
@@ -29,7 +29,7 @@ fi
 # Backup store log just in case
 _file="${logd}/smp-server-store.log"
 if [ -f "${_file}" ]; then
-  _backup_extension="$(date --universal '+%FT%T')"
+  _backup_extension="$(date --universal '+%04Y-%03m-%02dT%T')"
   cp -v -p "${_file}" "${_file}.${_backup_extension:-date-failed}"
   unset -v _backup_extension
 fi
--- a/scripts/docker/entrypoint-smp-server
+++ b/scripts/docker/entrypoint-smp-server
@@ -1,12 +1,12 @@
 #!/usr/bin/env sh
-confd="/etc/opt/simplex"
-logd="/var/opt/simplex/"
+confd='/etc/opt/simplex'
+logd='/var/opt/simplex/'
 
 # Check if server has been initialized
 if [ ! -f "$confd/smp-server.ini" ]; then
   # If not, determine ip or domain
   case "$ADDR" in
-    '') printf "Please specify \$ADDR environment variable.\n"; exit 1 ;;
+    '') printf 'Please specify $ADDR environment variable.\n'; exit 1 ;;
     *[a-zA-Z]*)
       case "$ADDR" in
         *:*) set -- --ip "$ADDR" ;;
--- a/scripts/docker/entrypoint-smp-server
+++ b/scripts/docker/entrypoint-smp-server
@@ -27,9 +27,16 @@ if [ ! -f "${confd}/smp-server.ini" ]; then
 fi
 
 # Backup store log just in case
+#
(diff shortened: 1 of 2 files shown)


shumvgolove:
Hey, could you rebase your changes against master? After that, I'll make a review.

tcely:
@shumvgolove Rebased and I also made similar changes for the new entrypoint file you added.

scripts/docker/entrypoint-smp-server
```diff
@@ -1,33 +1,39 @@
...
 
   # And init certificates and configs
   smp-server init -y -l "$@"
 fi
 
 # Backup store log just in case
-[ -f "$logd/smp-server-store.log" ] && cp "$logd"/smp-server-store.log "$logd"/smp-server-store.log."$(date +'%FT%T')"
+_file="${logd}/smp-server-store.log"
+if [ -f "${_file}" ]; then
+  _backup_extension="$(date --universal '+%04Y-%03m-%02dT%T')"
```
shumvgolove, line 32: - `--universal` is [not POSIX complaint](https://pubs.opengroup.org/onlinepubs/009695399/utilities/date.html). You should change it to `-u` instead.
- What's the point of rewriting `+'%FT%T'` to `'+%04Y-%03m-%02dT%T'`?

   > Both YYYY-MM-DD and YYYY-DD-MM are used by people and can be confusing in the beginning of the month.
   
   Doesn't seem like a valid reason.  Current behavior is just a `ISO 8601` format without `TZ`.
- Please specify case, where `date` command can fail (this is continuation of https://github.com/simplex-chat/simplexmq/pull/746#issuecomment-1536316845). Otherwise, I see no point of changing default behavior (without `_backup_extension` and `unset`'ting it).

shumvgolove (changes requested):

scripts/docker/entrypoint-smp-server
```diff
@@ -1,33 +1,39 @@
 #!/usr/bin/env sh
-confd="/etc/opt/simplex"
-logd="/var/opt/simplex/"
+confd='/etc/opt/simplex'
+logd='/var/opt/simplex/'
 
 # Check if server has been initialized
-if [ ! -f "$confd/smp-server.ini" ]; then
+if [ ! -f "${confd}/smp-server.ini" ]; then
```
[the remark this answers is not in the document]
tcely: The point of adding the braces is to reduce cognitive load on the people reading the script and to make the intention clearer.

The user, in my view, is the person evaluating your docker image, who may decide to deploy it or not.

scripts/docker/entrypoint-smp-server
tcely: > * `--universal` is [not POSIX complaint](https://pubs.opengroup.org/onlinepubs/009695399/utilities/date.html). You should change it to `-u` instead.

That's a fair point. I'll change that and add a comment to maintain the clarity advantage.

> Doesn't seem like a valid reason. Current behavior is just a ISO 8601 format without TZ.

That isn't actually communicated though. Think about the person who notices these files showing up after using this image. How do they know what the format is? Do you want to force them to stat the file to guess at what the format might be?

I'll add a comment to the entrypoint block explaining this now that I have brought it up. I like the format that doesn't make me guess which number is the month and which time zone was used much better if all I see is the file names. This is what `docker diff` shows, for example.

> Otherwise, I see no point of changing default behavior (without _backup_extension and unset'ting it).

This is another case of what the user sees being more important than how you feel about the code looking. (Even though I prefer not hiding a command substitution at the end of a long line, either way.)

`date` can fail for a bad flag, a changed path, a different interpreter or libc conflict, etc. A user could mount over it from the host, you can't prevent things like this. How it fails is beside the point, for me.

I don't see a reason to prefer `file.log.` over the much more informative `file.log.date-failed`, especially when that change costs nothing in the working case.

shumvgolove: > That isn't actually communicated though.

Agree. If you feel like this would be worthwhile change, then please go ahead and add this comment to the script, while keeping the international date format in place. Because you're "like the format" with leading "0" in Month isn't valid nor constructive.

> Even though I prefer not hiding a command substitution at the end of a long line, either way.

Actually, I agree with that too, but:

> date can fail for a bad flag, a changed path, a different interpreter or libc conflict, etc. A user could mount over it from the host, you can't prevent things like this. How it fails is beside the point, for

A user purposely sabotaging the docker container isn't something we can prevent. I'll give it a pass since this isn't really a huge change.

tcely: I put the ISO 8601 format without the time zone back.

That standard and how it was modified is also described in the comments.

tcely:
@shumvgolove GitHub did not seem to track this PR state correctly.

The date format has been reverted and appropriate comments are included.

shumvgolove (approved):

shumvgolove:
Thanks!

PR #764, opened 2023-05-31
connection switch - track status, stop switch command

spaced4ndy:


src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -473,19 +477,21 @@ instance ToJSON QueueDirection where
...
-data SwitchPhase = SPStarted | SPConfirmed | SPCompleted
+data SwitchPhase = SPStarted | SPConfirmed | SPFinalizing | SPCompleted
   deriving (Eq, Show)
 
 instance StrEncoding SwitchPhase where
   strEncode = \case
     SPStarted -> "started"
     SPConfirmed -> "confirmed"
     SPCompleted -> "completed"
+    SPFinalizing -> "finalizing"
```
epoberezkin, line 488: different order from type

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -495,19 +501,147 @@ instance ToJSON SwitchPhase where
 instance FromJSON SwitchPhase where
   parseJSON = strParseJSON "SwitchPhase"
 
+data RcvSwitchStatus
+  = -- set in beginning of switchConnectionAsync' before queueing SWCH command
+    RSSQueueingSwch
```
epoberezkin, line 506: suggest shortening namespace to RS/SS or even R/S - it won't be less clear.

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -495,19 +501,147 @@ instance ToJSON SwitchPhase where
...
+instance FromJSON SndSwitchStatus where
+  parseJSON = strParseJSON "SndSwitchStatus"
+
+data RcvQueueInfo = RcvQueueInfo
+  { rcvServer :: SMPServer,
+    rcvSwitchStatus :: Maybe RcvSwitchStatus
+  }
+  deriving (Eq, Show, Generic)
+
+instance ToJSON RcvQueueInfo where toEncoding = J.genericToEncoding J.defaultOptions
```
epoberezkin, line 606: we usually use modified options when there is Maybe to not include nulls.

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -80,6 +82,13 @@ data RcvQueue = RcvQueue
   }
   deriving (Eq, Show)
 
+rcvQueueInfo :: RcvQueue -> RcvQueueInfo
+rcvQueueInfo q@RcvQueue {switchStatus} =
+  RcvQueueInfo
+    { rcvServer = qServer q,
```
epoberezkin, line 88:
```suggestion
rcvQueueInfo RcvQueue {server, switchStatus} =
  RcvQueueInfo
    { rcvServer = server,
```

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -113,11 +122,19 @@ data SndQueue = SndQueue
...
+    switchStatus :: Maybe SndSwitchStatus,
     -- | SMP client version
     smpClientVersion :: Version
   }
   deriving (Eq, Show)
 
+sndQueueInfo :: SndQueue -> SndQueueInfo
+sndQueueInfo q@SndQueue {switchStatus} =
+  SndQueueInfo
+    { sndServer = qServer q,
```
epoberezkin, line 134:
```suggestion
sndQueueInfo SndQueue {server, switchStatus} =
  SndQueueInfo
    { sndServer = server,
```

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -155,6 +172,14 @@ findRQ :: (SMPServer, SMP.SenderId) -> NonEmpty RcvQueue -> Maybe RcvQueue
 findRQ sAddr = find $ sameQAddress sAddr . sndAddress
 {-# INLINE findRQ #-}
 
+findSwitchedRQ :: NonEmpty RcvQueue -> Maybe RcvQueue
```
epoberezkin, line 175: switching? what if there is more than one?

src/Simplex/Messaging/Agent/Store.hs
```diff
@@ -155,6 +172,14 @@ findRQ :: (SMPServer, SMP.SenderId) -> NonEmpty RcvQueue -> Maybe RcvQueue
...
+findSwitchedRQ = find $ isJust . (switchStatus :: RcvQueue -> Maybe RcvSwitchStatus)
+{-# INLINE findSwitchedRQ #-}
+
+findSwitchedSQ :: NonEmpty SndQueue -> Maybe SndQueue
+findSwitchedSQ = find $ isJust . (switchStatus :: SndQueue -> Maybe SndSwitchStatus)
```
epoberezkin, line 180: used only once? maybe ok to keep it.

src/Simplex/Messaging/Agent.hs
```diff
@@ -527,7 +532,14 @@ deleteConnectionsAsync_ onSuccess c connIds = case connIds of
 switchConnectionAsync' :: AgentMonad m => AgentClient -> ACorrId -> ConnId -> m ()
 switchConnectionAsync' c corrId connId =
   withStore c (`getConn` connId) >>= \case
-    SomeConn _ DuplexConnection {} -> enqueueCommand c corrId connId Nothing $ AClientCommand $ APC SAEConn SWCH
+    SomeConn _ (DuplexConnection _ rqs@(rq :| _rqs) _) -> case findSwitchedRQ rqs of
+      Nothing -> do
+        let queueNotSwitched = \RcvQueue {switchStatus} -> isNothing switchStatus
```
epoberezkin, line 537: this check seems redundant. You tried to find the queue that switches, found nothing, which means that the primary queue is also not switching - why check it separately?

src/Simplex/Messaging/Agent.hs
```diff
@@ -1979,6 +2134,48 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
+      _ -> pure $ Right (Left QCEQueueNotFound)
+
+data QueueCheckError = QCEQueueNotFound | QCEPredicateFailed
+
+qceStr :: QueueCheckError -> String -> String
+qceStr qce qpFailedStr = case qce of
+  QCEQueueNotFound -> "queue not found"
+  QCEPredicateFailed -> qpFailedStr
+
+switchCannotProceedErr :: AgentMonad m => String -> m a
```
epoberezkin, line 2176: very verbose

src/Simplex/Messaging/Agent.hs
```diff
@@ -1979,6 +2134,48 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
+switchCannotProceedErr errStr = throwError $ AGENT $ A_QUEUE $ "cannot proceed with switch, " <> errStr
```
epoberezkin, line 2177:
```suggestion
switchCannotProceedErr errStr = throwError $ AGENT $ A_QUEUE $ "switch error: " <> errStr
```

src/Simplex/Messaging/Agent.hs
epoberezkin, line 2176:
```suggestion
switchErr :: AgentMonad m => String -> m a
```

src/Simplex/Messaging/Agent.hs
```diff
@@ -527,7 +532,14 @@ deleteConnectionsAsync_ onSuccess c connIds = case connIds of
...
 switchConnectionAsync' c corrId connId =
   withStore c (`getConn` connId) >>= \case
-    SomeConn _ DuplexConnection {} -> enqueueCommand c corrId connId Nothing $ AClientCommand $ APC SAEConn SWCH
+    SomeConn _ (DuplexConnection _ rqs@(rq :| _rqs) _) -> case findSwitchedRQ rqs of
+      Nothing -> do
+        let queueNotSwitched = \RcvQueue {switchStatus} -> isNothing switchStatus
+            updateSwitchStatus = \db q -> setRcvQueueSwitchStatus db q (Just RSSQueueingSwch)
+        withExistingRQ c rq queueNotSwitched updateSwitchStatus >>= \case
+          Right _replaced' -> enqueueCommand c corrId connId Nothing $ AClientCommand $ APC SAEConn SWCH
+          Left qce -> switchCannotProceedErr $ qceStr qce "queue expected to be not switched"
```
epoberezkin, line 541: this is just store operation, no network IO - why can't we just update the status after it's been enqueued and do it in a single transaction? the status then will be RSSwchQueued

src/Simplex/Messaging/Agent.hs
```diff
@@ -1979,6 +2134,48 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
           | internalPrevMsgHash /= receivedPrevMsgHash = MsgError MsgBadHash
           | otherwise = MsgError MsgDuplicate -- this case is not possible
 
+withSwitchedRQ :: AgentMonad m => AgentClient -> RcvQueue -> RcvSwitchStatus -> (DB.Connection -> RcvQueue -> IO a) -> m (Either QueueCheckError a)
+withSwitchedRQ c rq expected =
+  withExistingRQ c rq (\RcvQueue {switchStatus} -> switchStatus == Just expected)
+
+withExistingRQ :: AgentMonad m => AgentClient -> RcvQueue -> (RcvQueue -> Bool) -> (DB.Connection -> RcvQueue -> IO a) -> m (Either QueueCheckError a)
+withExistingRQ c RcvQueue {connId, dbQueueId} qp action =
```
epoberezkin, line 2142: why are we reading it here again if we already have it? should it just be used?

epoberezkin: this function seems completely redundant - it reads what it already have and then performs the passed action. All that's needed is to call the action from the places where it's called, and this function can be then removed.

src/Simplex/Messaging/Agent.hs
```diff
@@ -1979,6 +2134,48 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
+            pure $ Right (Right r)
+          else pure $ Right (Left QCEPredicateFailed)
+      _ -> pure $ Right (Left QCEQueueNotFound)
+
+withSwitchedSQ :: AgentMonad m => AgentClient -> SndQueue -> SndSwitchStatus -> (DB.Connection -> SndQueue -> IO a) -> m (Either QueueCheckError a)
+withSwitchedSQ c sq expected =
+  withExistingSQ c sq (\SndQueue {switchStatus} -> switchStatus == Just expected)
+
+withExistingSQ :: AgentMonad m => AgentClient -> SndQueue -> (SndQueue -> Bool) -> (DB.Connection -> SndQueue -> IO a) -> m (Either QueueCheckError a)
+withExistingSQ c SndQueue {connId, dbQueueId} qp action =
```
epoberezkin, line 2158: same here, just remove this function - there is no need to read the queue again.

epoberezkin (commented):
general feedback it's much more complex than it needs to be, with some unnecessary indirections and repetitions. some statuses also seem unnecessary as they are not used

src/Simplex/Messaging/Agent.hs
```diff
@@ -867,7 +879,16 @@ runCommandProcessing c@AgentClient {subQ} server_ = do
             notify OK
         LET confId ownCInfo -> withServer' . tryCommand $ allowConnection' c connId confId ownCInfo >> notify OK
         ACK msgId -> withServer' . tryCommand $ ackMessage' c connId msgId >> notify OK
-        SWCH -> noServer $ tryCommand $ switchConnection' c connId >>= notify . SWITCH QDRcv SPStarted
+        SWCH ->
+          noServer $
+            tryCommand $
+              withConnLock c connId "switchConnection" $
```
epoberezkin, line 885: they read better on one line

epoberezkin, line 885:
```suggestion
          noServer . tryCommand . withConnLock c connId "switchConnection" $
```

src/Simplex/Messaging/Agent.hs
```diff
@@ -1178,21 +1229,80 @@ ackMessage' c connId msgId = withConnLock c connId "ackMessage" $ do
...
-      srv' <- if srv == server then getNextServer c userId [server] else pure srvAuth
-      (q, qUri) <- newRcvQueue c userId connId srv' clientVRange
-      let rq' = (q :: RcvQueue) {primary = True, dbReplaceQueueId = Just dbQueueId}
-      void . withStore c $ \db -> addConnRcvQueue db connId rq'
+switchConnection' c connId =
+  withConnLock c connId "switchConnection" $
+    withStore c (`getConn` connId) >>= \case
+      SomeConn _ conn@(DuplexConnection _ rqs@(replaced :| rqs_) _) -> case findSwitchedRQ rqs of
+        Nothing -> do
+          let queueNotSwitched = \RcvQueue {switchStatus} -> isNothing switchStatus
```
epoberezkin, line 1237: same comment - this is always true here.

src/Simplex/Messaging/Agent.hs
```diff
@@ -887,28 +908,51 @@ runCommandProcessing c@AgentClient {subQ} server_ = do
...
+                            getRcvQueuesByConnId db connId
+                      withSwitchedRQ c replaced' RSSSecureStarted updateSwitchStatus >>= \case
+                        Right (Just rqs') -> do
+                          void . enqueueMessages c cData sqs SMP.noMsgFlags $ QUSE [((server, sndId), True)]
+                          let conn' = DuplexConnection cData rqs' sqs
+                          notify . SWITCH QDRcv SPFinalizing $ connectionStats conn'
+                        Right _ -> throwError $ INTERNAL "no rcv queues in connection after processing ICQSecure"
+                        Left qce -> switchCannotProceedErr $ qceStr qce ("expected switch status " <> show RSSSecureStarted)
+                    _ -> internalErr "ICQSecure: queue address not found in connection"
+              Nothing -> internalErr "ICQSecure: no switched queue found"
```
epoberezkin, line 932: very large function, completely unclear what's going on here. Refactor/simplify

src/Simplex/Messaging/Agent.hs
```diff
@@ -1901,40 +2013,69 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
-                  void . withStore c $ \db -> addConnSndQueue db connId sq'
-                  case (sndPublicKey, e2ePubKey) of
-                    (Just sndPubKey, Just dhPublicKey) -> do
-                      logServer "<--" c srv rId $ "MSG <QADD> " <> logSecret (senderId queueAddress)
-                      let sqInfo' = (sqInfo :: SMPQueueInfo) {queueAddress = queueAddress {dhPublicKey}}
-                      void . enqueueMessages c cData sqs SMP.noMsgFlags $ QKEY [(sqInfo', sndPubKey)]
-                      let conn' = DuplexConnection cData rqs (sq <| sq' :| sqs_)
-                      notify . SWITCH QDSnd SPStarted $ connectionStats conn'
-                    _ -> qError "absent sender keys"
+                (_, Just replaced@SndQueue {dbQueueId}) -> do
```
epoberezkin, line 2023: also very large and complex - needs to be broken down into separate functions so it's clear what's going on, unnecessary checks and database reads removed, refactored and simplify.

tests/AgentTests/FunctionalAPITests.hs
```diff
@@ -125,6 +126,11 @@ runRight action =
     Right x -> pure x
     Left e -> error $ "Unexpected error: " <> show e
 
+shouldContainPredicate :: (HasCallStack, Show a) => [a] -> (a -> Bool) -> Expectation
```
epoberezkin, line 129: It's only used once so it's not needed, just check it in place.

tests/AgentTests/FunctionalAPITests.hs
```diff
@@ -1037,6 +1063,139 @@ testSwitchDelete servers = do
...
+    subscribeConnection b aId
+
+    (_, _, r1) <- get a
+    (_, _, r2) <- get a
+    let rcvSwitchConfirmed = \case
+          SWITCH QDRcv SPConfirmed _ -> True
+          _ -> False
+    liftIO $ do
+      [r1, r2] `shouldContain` [ERR $ AGENT $ A_QUEUE "QKEY: queue address not found in connection"]
+      [r1, r2] `shouldContainPredicate` rcvSwitchConfirmed
```
epoberezkin, line 1143:
```suggestion
      rcvSwitchConfirmed r1 || rcvSwitchConfirmed r2 `shouldBe` True
```

src/Simplex/Messaging/Agent/Store.hs
spaced4ndy: I've used "switching" for queue that is being switched to. Currently it's not possible to have more than one switched queue, because when switching connection we check there's no already switched queue. For redundancy it would have to be adjusted somehow, but also api might change as well from switching connection to possibly switching specific queues

src/Simplex/Messaging/Agent/Store.hs
spaced4ndy: yes

src/Simplex/Messaging/Agent.hs
spaced4ndy: the find is outside of transaction that updates queue status - the second check may be unnecessary because of conn lock, but I thought to keep it as a failsafe

src/Simplex/Messaging/Agent.hs
spaced4ndy: to prohibit queueing second SWCH command or simultaneously calling synchronous function, though it may be a bit redundant

src/Simplex/Messaging/Agent.hs
spaced4ndy: it re-reads the queue to check there were no concurrent database changes. Again, may be unnecessary due to conn lock but I kept it as a separate fail-safe

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -473,18 +477,20 @@ instance ToJSON QueueDirection where
 instance FromJSON QueueDirection where
   parseJSON = strParseJSON "QueueDirection"
 
-data SwitchPhase = SPStarted | SPConfirmed | SPCompleted
+data SwitchPhase = SPStarted | SPConfirmed | SPFinalizing | SPCompleted
```
epoberezkin, line 480: SPFinalizing - not clear what it actually means... Need to check when exactly it happens.

src/Simplex/Messaging/Agent/Store.hs
epoberezkin: > queue that is being switched to

where is that? It's unclear that "switching" means "switched to". Why not just call it "switching to"?

src/Simplex/Messaging/Agent.hs
```diff
@@ -526,9 +531,14 @@ deleteConnectionsAsync_ onSuccess c connIds = case connIds of
...
-  withStore c (`getConn` connId) >>= \case
-    SomeConn _ DuplexConnection {} -> enqueueCommand c corrId connId Nothing $ AClientCommand $ APC SAEConn SWCH
-    _ -> throwError $ CMD PROHIBITED
+  withConnLock c connId "switchConnectionAsync" $
+    withStore c (`getConn` connId) >>= \case
+      SomeConn _ (DuplexConnection _ rqs@(rq :| _rqs) _) -> case findSwitchedRQ rqs of
+        Nothing -> do
+          void $ withStore' c $ \db -> setRcvQueueSwitchStatus db rq (Just RSSwitchStarted)
+          enqueueCommand c corrId connId Nothing $ AClientCommand $ APC SAEConn SWCH
+        Just _ -> throwError $ AGENT $ A_QUEUE "connection already switching"
```
epoberezkin, line 540: swap branches

epoberezkin: you should never change the order of predicates based on the contents, it's just confusing. Just should always come first, even if it's just throw.

epoberezkin: the reasoning that "what I do in Nothing" is the main behaviour, so let's put it first, is flawed. Think like you make a control flow diagram - you alway have consistent positioning of True/False branches there, it shouldn't depend on what's inside.

src/Simplex/Messaging/Agent.hs
```diff
@@ -889,26 +904,47 @@ runCommandProcessing c@AgentClient {subQ} server_ = do
...
+                    withStore' c $ \db -> setRcvQueueStatus db rq' Secured
+                    void . enqueueMessages c cData sqs SMP.noMsgFlags $ QUSE [((server, sndId), True)]
+                    rqs_ <- withStore' c $ \db -> do
+                      void $ setRcvQueueSwitchStatus db replaced (Just RSSendingQUSE)
+                      getRcvQueuesByConnId db connId
+                    case rqs_ of
+                      Just rqs' -> do
+                        let conn' = DuplexConnection cData rqs' sqs
+                        notify . SWITCH QDRcv SPFinalizing $ connectionStats conn'
+                      Nothing -> throwError $ INTERNAL "no rcv queues in connection after processing ICQSecure"
```
epoberezkin, line 925:
```suggestion
                      _ -> throwError $ INTERNAL "no rcv queues in connection after processing ICQSecure"
```

epoberezkin: house style is to never spell Nothing

src/Simplex/Messaging/Agent.hs
```diff
@@ -1125,18 +1162,19 @@ runSmpQueueMsgDelivery c@AgentClient {subQ} cData@ConnData {userId, connId, dupl
...
-                              -- remove the delivery from the map to stop the thread when the delivery loop is complete
-                              atomically $ TM.delete (qAddress sq') $ smpQueueMsgQueues c
-                              withStore' c $ \db -> do
-                                when primary $ setSndQueuePrimary db connId sq
-                                deletePendingMsgs db connId sq'
-                                deleteConnSndQueue db connId sq'
-                              let sqs'' = sq'' :| sqs'
-                                  conn' = DuplexConnection cData' rqs sqs''
-                              notify . SWITCH QDSnd SPCompleted $ connectionStats conn'
+                            Just (replaced, sq' : sqs') ->
```
epoberezkin, line 1167: swap branches

src/Simplex/Messaging/Agent.hs
```diff
@@ -1178,21 +1216,72 @@ ackMessage' c connId msgId = withConnLock c connId "ackMessage" $ do
...
-      pure . connectionStats $ DuplexConnection cData (rq <| rq' :| rqs_) sqs
-    _ -> throwError $ CMD PROHIBITED
+switchConnection' c connId =
+  withConnLock c connId "switchConnection" $
+    withStore c (`getConn` connId) >>= \case
+      SomeConn _ conn@(DuplexConnection _ rqs@(replaced :| _rqs) _) -> case findSwitchedRQ rqs of
+        Nothing -> do
+          replaced' <- withStore' c $ \db -> setRcvQueueSwitchStatus db replaced (Just RSSwitchStarted)
+          switchDuplexConnection c conn replaced'
+        Just _ -> throwError $ AGENT $ A_QUEUE "connection already switching"
```
epoberezkin, line 1226: swap branches

src/Simplex/Messaging/Agent.hs
```diff
@@ -889,26 +904,47 @@ runCommandProcessing c@AgentClient {subQ} server_ = do
...
-                    atomically $ sendNtfSubCommand ns (connId, NSCCreate)
-                  let conn' = DuplexConnection cData (rq'' :| rqs') sqs
-                  notify $ SWITCH QDRcv SPCompleted $ connectionStats conn'
+                | otherwise ->
+                  withSwitchedRQ replaced RSReceivedMessage $ do
+                    tryError (deleteQueue c replaced) >>= \case
+                      Left e
+                        | temporaryOrHostError e -> throwError e
+                        | otherwise -> finalizeSwitch replaced >> throwError e
+                      Right () -> finalizeSwitch replaced
```
epoberezkin, line 939: swap branches

src/Simplex/Messaging/Agent.hs
```diff
@@ -1979,6 +2102,28 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
           | internalPrevMsgHash /= receivedPrevMsgHash = MsgError MsgBadHash
           | otherwise = MsgError MsgDuplicate -- this case is not possible
 
+withSwitchedRQ :: AgentMonad m => RcvQueue -> RcvSwitchStatus -> m a -> m a
```
epoberezkin, line 2105: This is bad decomposition. It does two things - checks the condition, and then calls actions. What was needed is a function internalErr, and two lines in all places where it is called. And I am not even sure it should be INTERNAL error

epoberezkin: what was the point of passing action here and not just calling it outside?

src/Simplex/Messaging/Agent.hs
```diff
@@ -1178,21 +1216,72 @@ ackMessage' c connId msgId = withConnLock c connId "ackMessage" $ do
...
+switchDuplexConnection :: AgentMonad m => AgentClient -> Connection 'CDuplex -> RcvQueue -> m ConnectionStats
+switchDuplexConnection c (DuplexConnection cData@ConnData {connId, userId} rqs sqs) rq@RcvQueue {server, dbQueueId, sndId} = do
+  checkRQSwchStatus rq RSSwitchStarted
+  clientVRange <- asks $ smpClientVRange . config
+  -- try to get the server that is different from all queues, or at least from the primary rcv queue
+  srvAuth@(ProtoServerWithAuth srv _) <- getNextServer c userId $ map qServer (L.toList rqs) <> map qServer (L.toList sqs)
+  srv' <- if srv == server then getNextServer c userId [server] else pure srvAuth
+  (newQ, qUri) <- newRcvQueue c userId connId srv' clientVRange
+  let rq' = (newQ :: RcvQueue) {primary = True, dbReplaceQueueId = Just dbQueueId}
+  withStore' c $ \db -> void $ addConnRcvQueue db connId rq'
```
epoberezkin, line 1238: why was this change made? it seems before, it was a failable operation, and now even if addConnRcvQueue fails this doesn't?

src/Simplex/Messaging/Agent.hs
```diff
@@ -1901,40 +1992,61 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
-                      logServer "<--" c srv rId $ "MSG <QADD> " <> logSecret (senderId queueAddress)
-                      let sqInfo' = (sqInfo :: SMPQueueInfo) {queueAddress = queueAddress {dhPublicKey}}
-                      void . enqueueMessages c cData sqs SMP.noMsgFlags $ QKEY [(sqInfo', sndPubKey)]
-                      let conn' = DuplexConnection cData rqs (sq <| sq' :| sqs_)
-                      notify . SWITCH QDSnd SPStarted $ connectionStats conn'
-                    _ -> qError "absent sender keys"
+                (_, Just sq@SndQueue {dbQueueId}) -> do
+                  sq' <- withStore' c $ \db -> do
+                    switchingQueues <- getSwitchingSndQueues db dbQueueId
+                    forM_ switchingQueues $ \switchingQ -> deleteConnSndQueue db connId switchingQ
```
epoberezkin, line 2004: why we are reading queues again?
Also long and descriptive names for local variables is an anti-pattern - there are descriptive function names for that, local variables should be just letters, it makes it clearer they are local.

src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -1631,44 +1668,67 @@ rcvQueueQuery =
...
-      let server = SMPServer host port keyHash
-       in SndQueue {userId, connId, server, sndId, sndPublicKey, sndPrivateKey, e2ePubKey, e2eDhSecret, status, dbQueueId, primary, dbReplaceQueueId, smpClientVersion}
     primaryFirst SndQueue {primary = p, dbReplaceQueueId = i} SndQueue {primary = p', dbReplaceQueueId = i'} =
       -- the current primary queue is ordered first, the next primary - second
       compare (Down p) (Down p') <> compare i i'
 
+getSwitchingSndQueues :: DB.Connection -> Int64 -> IO [SndQueue]
+getSwitchingSndQueues db dbReplaceQueueId =
+  map toSndQueue
+    <$> DB.query db (sndQueueQuery <> "WHERE q.replace_snd_queue_id = ?") (Only dbReplaceQueueId)
```
epoberezkin, line 1700: I don't think it's correct - this is ID within connection, not global, same for another function

epoberezkin: It only works in tests because only one switch is in progress - pretty certain if you add several concurrent switches it'll fail. We should add tests like that anyway.

src/Simplex/Messaging/Agent.hs
```diff
@@ -1178,21 +1216,72 @@ ackMessage' c connId msgId = withConnLock c connId "ackMessage" $ do
...
+              (rqs_, switchingQueues) <- withStore' c $ \db -> do
+                -- multiple switching queues were possible when repeating switch while in progress was allowed
+                switchingQueues <- getSwitchingRcvQueues db $ dbQId replaced
+                forM_ switchingQueues $ \switchingQ -> deleteConnRcvQueue db switchingQ
+                void $ setRcvQueueSwitchStatus db replaced Nothing
+                rqs_ <- getRcvQueuesByConnId db connId
+                pure (rqs_, switchingQueues)
+              case rqs_ of
+                Just rqs' -> do
+                  forM_ switchingQueues $ \q -> deleteQueue c q `catchError` \_ -> pure ()
```
epoberezkin, line 1265: we need to make it asynchronous

src/Simplex/Messaging/Agent.hs
epoberezkin: also - why do we need to delete other queues on QADD?

src/Simplex/Messaging/Agent.hs
```diff
@@ -1178,21 +1216,72 @@ ackMessage' c connId msgId = withConnLock c connId "ackMessage" $ do
...
+                  let conn' = DuplexConnection cData rqs' sqs
+                  pure $ connectionStats conn'
+                _ -> throwError $ INTERNAL "no rcv queues in connection after switch stop"
+            else throwError $ AGENT $ A_QUEUE "switch stop error: switch cannot be stopped"
```
epoberezkin, line 1269: AGENT are errors of other agents, it is incorrect to use in this context

src/Simplex/Messaging/Agent.hs
```diff
@@ -1944,16 +2056,28 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
-            Just sq' -> do
-              logServer "<--" c srv rId $ "MSG <QUSE> " <> logSecret (snd addr)
-              withStore' c $ \db -> setSndQueueStatus db sq' Secured
-              let sq'' = (sq' :: SndQueue) {status = Secured}
-              -- sending QTEST to the new queue only, the old one will be removed if sent successfully
-              void $ enqueueMessages c cData [sq''] SMP.noMsgFlags $ QTEST [addr]
-              notify . SWITCH QDSnd SPConfirmed $ connectionStats conn
-            _ -> qError "QUSE: queue address not found in connection"
+        qUseMsg ((addr, _primary) :| _) (DuplexConnection _ rqs sqs) =
+          case switchingSQ sqs of
```
epoberezkin, line 2060: searching for the switching queue seems a wrong approach, it must be determined by the context.

There are two options:
1. only primary queue can be switched - in this case it's always the first queue, and we just need to check it's switching status, not search for this queue. It also creates a potential vulnerability, as the message can arrive to another queue - it should be just ignored in this case.
2. we want data model that supports redundancy and multiple queues can be switched concurrently.

In either case, this should not be determined by queue switching status, but should be contextual.

src/Simplex/Messaging/Agent.hs
```diff
@@ -1944,16 +2056,28 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
+            Just replaced -> do
+              checkSQSwchStatus replaced SSSendingQKEY
+              replaced' <- withStore' c $ \db -> setSndQueueSwitchStatus db replaced (Just SSReceivedQUSE)
```
epoberezkin, line 2063: why do we need the status in the beginning of the operation?

it's never checked for, not sent to the client, and if the connection stays in it it certainly can't progress switching. 

There are other cases like that too.

src/Simplex/Messaging/Agent.hs
spaced4ndy: we delete other queues on QADD to cleanup queues that haven't completed switch, for example because sender stopped it, or to cleanup extra queues from multiple simultaneous switches in current version

src/Simplex/Messaging/Agent.hs
```diff
@@ -887,13 +902,24 @@ runCommandProcessing c@AgentClient {subQ} server_ = do
...
+          withStore' c (`deleteConnRcvQueue` rq)
         ICQSecure rId senderKey ->
           withServer $ \srv -> tryWithLock "ICQSecure" . withDuplexConn $ \(DuplexConnection cData rqs sqs) ->
             case find (sameQueue (srv, rId)) rqs of
-              Just rq'@RcvQueue {server, sndId, status} -> when (status == Confirmed) $ do
-                secureQueue c rq' senderKey
-                withStore' c $ \db -> setRcvQueueStatus db rq' Secured
-                void . enqueueMessages c cData sqs SMP.noMsgFlags $ QUSE [((server, sndId), True)]
+              Just rq'@RcvQueue {server, sndId, status, dbReplaceQueueId = Just replaceQId} ->
+                case find (\RcvQueue {dbQueueId} -> dbQueueId == replaceQId) rqs of
```
epoberezkin, line 913:
```suggestion
                case find ((replaceQId ==) . dbQId) rqs of
```

src/Simplex/Messaging/Agent.hs
```diff
@@ -1178,21 +1214,67 @@ ackMessage' c connId msgId = withConnLock c connId "ackMessage" $ do
...
+  pure . connectionStats $ DuplexConnection cData rqs' sqs
+
+stopConnectionSwitch' :: AgentMonad m => AgentClient -> ConnId -> m ConnectionStats
+stopConnectionSwitch' c connId =
+  withConnLock c connId "stopConnectionSwitch" $
+    withStore c (`getConn` connId) >>= \case
+      SomeConn _ (DuplexConnection cData rqs sqs) -> case switchingRQ rqs of
+        Just rq -> do
+          if canStopRcvSwitch rq
+            then do
```
epoberezkin, line 1250: use guards

src/Simplex/Messaging/Agent.hs
```diff
@@ -1944,16 +2038,23 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
-            Just sq' -> do
-              logServer "<--" c srv rId $ "MSG <QUSE> " <> logSecret (snd addr)
-              withStore' c $ \db -> setSndQueueStatus db sq' Secured
-              let sq'' = (sq' :: SndQueue) {status = Secured}
-              -- sending QTEST to the new queue only, the old one will be removed if sent successfully
-              void $ enqueueMessages c cData [sq''] SMP.noMsgFlags $ QTEST [addr]
-              notify . SWITCH QDSnd SPConfirmed $ connectionStats conn
-            _ -> qError "QUSE: queue address not found in connection"
+            Just sq'@SndQueue {dbReplaceQueueId = Just replaceQId} -> do
+              case find (\SndQueue {dbQueueId} -> dbQueueId == replaceQId) sqs of
```
epoberezkin, line 2044:
```suggestion
              case find ((replaceQId ==) . dbQId) sqs of
```

PR #778, opened 2023-06-22
batch subscriptions for ntf server

epoberezkin:


src/Simplex/Messaging/Client/Agent.hs
```diff
@@ -201,27 +208,47 @@ getSMPServerClient' ca@SMPClientAgent {agentCfg, smpClients, msgQ} srv =
...
+        isNotifier = \case
+          SPNotifier -> True
+          SPRecipient -> False
+
+        subscribe_ :: SMPClient -> SMPSubParty -> [(SMPSub, C.APrivateSignKey)] -> IO [Either SMPClientError ()]
+        subscribe_ smp party subs = do
+          subs' <- filterM (atomically . hasSub (srvSubs ca) srv . fst) subs
+          case L.nonEmpty subs' of
+            Just subs'' -> do
+              let subs3 = L.map (first snd) subs''
```
spaced4ndy, line 230: hard to read.. maybe call it qIdsKeys and map with lambda

src/Simplex/Messaging/Notifications/Server.hs
```diff
@@ -154,11 +160,15 @@ ntfSubscriber NtfSubscriber {smpSubscribers, newSubQ, smpAgent = ca@SMPClientAge
...
     subscribe :: M ()
     subscribe =
-      forever $
-        atomically (readTBQueue newSubQ) >>= \case
-          sub@(NtfSub NtfSubData {smpQueue = SMPQueueNtf {smpServer}}) -> do
-            SMPSubscriber {newSubQ = subscriberSubQ} <- getSMPSubscriber smpServer
-            atomically $ writeTQueue subscriberSubQ sub
+      forever $ do
+        subs <- atomically (readTBQueue newSubQ)
+        let ss = L.groupBy ((==) `on` server) subs
```
spaced4ndy, line 165: why groupBy is needed both here and in `resubscribe` - is it a sanity check?

src/Simplex/Messaging/Client/Agent.hs
epoberezkin: that's ok I think - if your read it as change the "first" part to its "snd" part, where short form means "taking" the part, and long form means "changing" the part in place. Can operate on things like tuples or Either...

src/Simplex/Messaging/Notifications/Server.hs
epoberezkin: they operate on slightly different types, the top one on `NonEmpty NtfSubData`, the bottom one on `NonEmpty (NtfEntityRec 'Subscription)`, where the latter type is a wrapper around `NtfSubData`

Also they are called in different cases - the first when the server just starts, and the latter when ntfSubscriber receives the new subscriptions...

But maybe the latter is indeed sufficient and the former can just dump there the whole array without splitting by server? Need to think.

epoberezkin: The only difference is the delay - initially it was after each subscription, not sure if it was helpful or not, now it's between groups of subscriptions on different servers, and leaving only two would remove the delay completely...

epoberezkin: I don't remember why this delay was added in the first place?

PR #788, opened 2023-06-29
increase connection agent version according to received messages; add version and flags to connection stats, bump currentSMPAgentVersion

spaced4ndy:


src/Simplex/Messaging/Agent.hs
```diff
@@ -1293,8 +1294,8 @@ abortConnectionSwitch' c connId =
 synchronizeRatchet' :: AgentMonad m => AgentClient -> ConnId -> Bool -> m ConnectionStats
 synchronizeRatchet' c connId force = withConnLock c connId "synchronizeRatchet" $ do
   withStore c (`getConn` connId) >>= \case
-    SomeConn _ (DuplexConnection cData@ConnData {ratchetSyncState} rqs sqs)
-      | ratchetSyncState `elem` ([RSAllowed, RSRequired] :: [RatchetSyncState]) || force -> do
+    SomeConn _ (DuplexConnection cData rqs sqs)
+      | ratchetSyncAllowed' cData || force -> do
```
epoberezkin, line 1298: why don't we check here it's supported?

src/Simplex/Messaging/Agent.hs
```diff
@@ -1445,16 +1453,26 @@ getConnectionRatchetAdHash' c connId = do
...
+  NewConnection cData ->
+    stats cData
+  where
+    stats cData@ConnData {connAgentVersion, ratchetSyncState} =
+      ConnectionStats
+        { connAgentVersion,
+          rcvQueuesInfo = [],
+          sndQueuesInfo = [],
+          ratchetSyncState,
+          ratchetSyncAllowed = ratchetSyncAllowed' cData,
```
epoberezkin, line 1473: I thought the idea is to return True here even if it's not supported, so that in this case the button would show and suggest to ask another party to update. And in this case where required you need to check for both allowed and supported.

src/Simplex/Messaging/Agent.hs
epoberezkin: see comment below

src/Simplex/Messaging/Agent.hs
```diff
@@ -1912,6 +1935,18 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
                               _ -> pure Nothing
                       _ -> prohibited >> ack
                   _ -> prohibited >> ack
+              updateConnVersion :: Connection c -> ConnData -> Version -> m (Connection c, ConnData)
+              updateConnVersion conn' cData' agentVersion = do
+                aVRange <- asks $ smpAgentVRange . config
+                case versionToRange agentVersion `compatibleVersion` aVRange of
+                  Just _
```
epoberezkin, line 1942: you actually need to use the result of compatibleVersion to update the version, you can't just use `agentVersion` as it can be higher than compatible.

```suggestion
                  Just (Compatible v)
```

src/Simplex/Messaging/Agent.hs
```diff
@@ -1912,6 +1935,18 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
+                    | agentVersion > connAgentVersion -> do
+                      withStore' c $ \db -> setConnAgentVersion db connId agentVersion
+                      let cData'' = cData' {connAgentVersion = agentVersion} :: ConnData
```
epoberezkin, line 1945: you actually need to use the result of compatibleVersion to update the version, you can't just use agentVersion as it can be higher than compatible.

```suggestion
                case versionToRange agentVersion `compatibleVersion` aVRange of
                  Just (Compatible v)
                    | v > connAgentVersion -> do
                      withStore' c $ \db -> setConnAgentVersion db connId v
                      let cData'' = cData' {connAgentVersion = v} :: ConnData
```

epoberezkin: e.g. if connection is set to 2, your supported range is 1-3 and you receive 4, you want to increase to 3 not to 4 - that's what compatibleVersion determines.

Now I am thinking that maybe it's incorrect to use compatibleVersion at all, as it won't actually return 3 in this case, and to make it do it you need to pass range (minVersion aVRange, agentVersion)?

src/Simplex/Messaging/Version.hs
```diff
@@ -7,7 +7,7 @@
 
 module Simplex.Messaging.Version
   ( Version,
-    VersionRange (minVersion, maxVersion),
+    VersionRange (..),
```
epoberezkin, line 10: it was not exported for a reason, there is a safe smart constructor mkVersionRange to use outside of this module. You should not be able to construct VRange outside of this module

src/Simplex/Messaging/Agent.hs
```diff
@@ -1913,6 +1923,18 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
                               _ -> pure Nothing
                       _ -> prohibited >> ack
                   _ -> prohibited >> ack
+              updateConnVersion :: Connection c -> ConnData -> Version -> m (Connection c, ConnData)
```
epoberezkin, line 1926:
```suggestion
              updateConnVersion :: Connection c -> ConnData -> Version -> m (Connection c)
```

why not just return this?

src/Simplex/Messaging/Agent.hs
```diff
@@ -1913,6 +1923,19 @@ processSMPTransmission c@AgentClient {smpClients, subQ} (tSess@(_, srv, _), v, s
...
+              updateConnVersion conn' cData' msgAgentVersion = do
+                aVRange <- asks $ smpAgentVRange . config
+                let msgAVRange = fromMaybe (versionToRange msgAgentVersion) $ safeVersionRange (minVersion aVRange) msgAgentVersion
+                case msgAVRange `compatibleVersion` aVRange of
+                  Just (Compatible av)
+                    | av > connAgentVersion -> do
+                      withStore' c $ \db -> setConnAgentVersion db connId av
+                      let cData'' = cData' {connAgentVersion = av} :: ConnData
+                          conn'' = updateConnection cData'' conn'
+                      pure conn''
```
epoberezkin, line 1936:
```suggestion
                      pure updateConnection cData'' conn'
```

PR #805, opened 2023-07-18
server: control port state introspection, suspend/resume

epoberezkin:


epoberezkin:
done in another PR with 9.6 thread labelling.

PR #814, opened 2023-08-01
Fix stack build

dpwiz:
The original is issue is reported in the [chat repo](https://github.com/simplex-chat/simplex-chat/issues/2749).

### DIFF
--- a/src/Simplex/Messaging/Agent.hs
+++ b/src/Simplex/Messaging/Agent.hs
@@ -100,7 +100,7 @@ module Simplex.Messaging.Agent
   )
 where
 
-import Control.Concurrent.STM (stateTVar)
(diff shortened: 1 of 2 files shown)


epoberezkin:
I think we should simply remove stack yaml, tbh, the only reason we still run stack is for package.yaml/hpack - we just need to use hpack separately, but we never came around to figuring it out.

dpwiz:
Hm. Cabal build is currently broken on `master` due to dependency solver problems. Having a stack "backup" can be handy until the situation resolves. And it is possible to generate `cabal.project.freeze` files from it to prevent accidental breakage.

I can add a notice that the stack build is maintained by volunteers and may need fixes once in a while.

dpwiz:
Oops, apparently not.

PR #815, opened 2023-08-02
Pin GHC version in cabal.project

dpwiz:
Allows building without switching user-preferred version.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -1,3 +1,5 @@
+with-compiler: ghc-8.10.7
+
 packages: .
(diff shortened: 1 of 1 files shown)


epoberezkin:
this should be done in package.yaml somehow I think, so when processed with hpack it's retained.

dpwiz:
I don't think package has anything on it. You can pin the `base` version, but it would just fail to build instead of selecting a correct compiler from the range of that's available.

epoberezkin:
right, it overwrites .cabal inded.

PR #818, opened 2023-08-08
Upgrade to ghc 9.4.5

JonathanLorimer:
I found a working package set, and could update the `stack.yaml` to reflect these new packages.

One issue is that GHC 9.4.1 and greater requires that you disambiguate duplicate record fields (see bottom of [docs here](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/disambiguate_record_fields.html#disambiguate-fields))

I started disambiguating record fields by hand, but it was pretty invasive and manual. I was wondering if you would be open to turning on `-WAmbiguous-fields` and then I could work on slowly disambiguating them in several prs before coming back to this one.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -5,8 +5,8 @@ packages: .
 
 source-repository-package
     type: git
-    location: https://github.com/simplex-chat/aeson.git
-    tag: 3eb66f9a68f103b5f1489382aad89f5712a64db7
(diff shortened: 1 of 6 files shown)


JonathanLorimer:
rebasing on #791 seems like it got things compiling, just working on getting stack working.

JonathanLorimer:
Closing in favour of #819

Issue #813, opened 2023-07-31
Support newer GHCs


turion:
It seems that through the dependency `memory == 0.15`, this package is restricted to GHC 8.10, which is very old. To make contributions easier, it would be great to support at least GHC 9.4.

PR #819, opened 2023-08-08
Upgrade to GHC 9.4.5

JonathanLorimer:
This would close https://github.com/simplex-chat/simplexmq/issues/813

Builds on angerman/ghc-9.6

Adds a `flake.nix` which provides a development shell with all required dependencies.


epoberezkin:
https://github.com/simplex-chat/simplexmq/tree/angerman/ghc-9.6

That is WIP to migrate to 9.6

JonathanLorimer:
> https://github.com/simplex-chat/simplexmq/tree/angerman/ghc-9.6
> 
> That is WIP to migrate to 9.6

Would it be okay to migrate to 9.4.5 as an interim step? I am testing with 9.6 locally.

epoberezkin:
> migrate to 9.4.5 as an interim step?

What problem does it solve? There is motivation to migrate to 9.6.2, I'd rather we don't migrate twice.

JonathanLorimer:
This PR specifically solves a couple of problems:
- adds nix shell for development, which provides a single command all-in-one dev environment
- adds hie.yaml file for HLS support
- updates to 9.4.5, which offers better support for tooling (like HLS), reduces the effort (or at least the diff) when upgrading to 9.6, theoretically provides quality of life improvements like faster compile times. Also, due to the nature of nix's Haskell package sets it was easier to upgrade to 9.4.5 than to find a nixpkgs snapshot that worked with 8.10.7 and the pinned deps.

I gather that stack is no longer used, but I also brought that back up to working order. There was an issue on the AUR forum about how updating simplex was broken because they were using stack to build the project, and the stack.yaml was not kept up to date.

I totally understand if you don't want to merge this as it (presumably) doesn't affect your workflow, but it may aid other people's, and possibly reduce friction to contributing. There is an open issue about getting the project onto ghc 9.x would help to contribute, so interest is at least non-zero.

epoberezkin:
> This PR specifically solves a couple of problems

By "problems" I meant "problems of the users" - nothing in this list is something that users care about, so we shouldn't spend time on it.

9.6(.2) will solve users' problems, at least potentially, so there is a benefit of migrating to it. Migrating in two steps is a much higher cost than migrating in one step, as it would require updating the whole build pipeline, the patches to GHC that make it work on mobile and 32 bit platforms, full regression testing of mobile apps, and longer time in beta versions each time we migrate.

The GHC versions earlier than 9.6 don't seem beneficial enough to justify the migration - so if there are some obstacles now we can continue using 8.10.7 until they are resolved.

I understand that the industry tradition is frequent migrations to the new versions of everything as soon as they become available, often to the detriment of productivity and stability. We don't do it. It seems that upgrading the compiler any more frequently than once every 12-24 months, absent bugs, security or serious performance issues, is counter-productive.

JonathanLorimer:
Got it, you can close then I guess.

I just figured that this was relatively low cost since the work is already done. 

From my perspective better dev tooling indirectly provides user value. Like the old adage about the lumberjack that sharpens his axe. Its similar to how introspecting thread state doesn't directly add value to messenger users, but improves developer's ability to debug concurrency issues and provide end user value.

PR #832, opened 2023-08-24
server: get RTS stats via control port

dpwiz:
* Adds `stats-rts` to control port to peek at current GC stats etc.
* Adds `rtsopts` runtime component to support that, and adjusting RTS options without rebuilding.
* ~~Adds `eventlog` runtime component to stream RTS events (and future Debug.Trace events and markers) into machine-readable log (it is opt-in and has an overhead minuscule enough to be permanently available).~~ UPD: `eventlog` is always-on in GHC 9.6.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -97,6 +97,7 @@ executables:
       - simplexmq
     ghc-options:
       - -threaded
+      - -rtsopts
 
   ntf-server:
     source-dirs: apps/ntf-server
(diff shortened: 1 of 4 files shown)


dpwiz:
Removed the `eventlog` flag, but this one now waits for `master-9` to get in.

epoberezkin:
this is done in another PR

PR #844, opened 2023-09-13
rfc: SMP/XFTP proxies

epoberezkin:


rfcs/2023-09-12-second-relays.md
```diff
@@ -0,0 +1,121 @@
...
+
+1. To avoid increasing the complexity of self-hosting, we should not create additional server types, and existing SMP and XFTP servers should be extended to provide the 1st step relay functions.
+
+2. SMP sending relay should not be able to observe queue addresses and their count on the destination relays.
+
+3. There must be no identifiers and cyphertext in common in outgoing and incoming traffic in 1st step relays.
+
+4. Traffic between the client and destination relays must be e2e encrypted, with MITM mitigated using server identity (certificate fingerprint), ideally without adding the additional fingerprint in destination relay address.
+
+5. SMP sending relays should implement retry logic and hold messages while they are delivered. They also should push "sent" confirmation to the client. Question: does it mean the third tick? another kind of the first tick (or some other symbol) in the UX?
```
spaced4ndy, line 74: third tick may be an overkill (there's even still possibility for "read" marker - too many unclear indications on a sent message), agent should probably report SENT to client after sending to the first relay, or maybe it could be a different status / event for debugging, for example SENT_RELAY

rfcs/2023-09-12-second-relays.md
```diff
@@ -0,0 +1,121 @@
...
+
+6. SMP sending relays may also increase utility of the platform by allowing:
+
+- schedule sending messages.
```
spaced4ndy, line 78: how would it work with double ratchet, integrity? encrypted in past and delivered out of order, then would have to be decrypted with an old key? if delivered out of order, wouldn't it be a skipped message?

PR #853, opened 2023-09-25
sql: set journal_mode to WAL

epoberezkin:


src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -365,7 +367,11 @@ migrateSchema st migrations confirmMigrations = do
     confirm err = confirmOrExit $ migrationErrorDescription err
     run ms = do
       let f = dbFilePath st
+          fWal = f <> "-wal"
+          fShm = f <> "-shm"
       copyFile f (f <> ".bak")
+      whenM (doesFileExist fWal) $ copyFile fWal (f <> ".bak-wal")
+      whenM (doesFileExist fShm) $ copyFile fShm (f <> ".bak-shm")
```
spaced4ndy, line 374:
```suggestion
      whenM (doesFileExist fWal) $ copyFile fWal (f <> ".wal.bak")
      whenM (doesFileExist fShm) $ copyFile fShm (f <> ".shm.bak")
```

epoberezkin: the point was that .bak file is also a valid database with these names

PR #857, opened 2023-09-28
add X509 cert and TLS credentials generator

dpwiz:


src/Simplex/Messaging/Crypto.hs
```diff
@@ -967,6 +998,54 @@ sign' (PrivateKeyEd448 pk k) msg = SignatureEd448 $ Ed448.sign pk k msg
...
+signX509 :: (ASN1Object o, Eq o, Show o) => APrivateSignKey -> o -> SignedExact o
+signX509 key = fst . objectToSignedExact f
+  where
+    f bytes =
+      ( signatureBytes $ sign key bytes,
+        signatureAlgorithmX509 key,
+        ()
+      )
+
+-- XXX: By now, 'KeyHash' really, really isn't...
```
epoberezkin, line 1013:
```suggestion
-- XXX: By now, 'KeyHash' really, really isn't...
```

PR #875, opened 2023-10-27
add runTransportWith

dpwiz:


src/Simplex/Messaging/Transport/Server.hs
```diff
@@ -61,23 +64,31 @@ serverTransportConfig TransportServerConfig {logTLSErrors} =
 --
 -- All accepted connections are passed to the passed function.
 runTransportServer :: forall c m. (Transport c, MonadUnliftIO m) => TMVar Bool -> ServiceName -> T.ServerParams -> TransportServerConfig -> (c -> m ()) -> m ()
-runTransportServer started port serverParams cfg server = do
+runTransportServer started port = runTransportServerWith (runTCPServer started port) (transportName (TProxy :: TProxy c))
+
+-- | Run a transport server with provided connection setup and handler.
+runTransportServerWith :: forall c m. (Transport c, MonadUnliftIO m) => ((Socket -> IO ()) -> IO ()) -> String -> T.ServerParams -> TransportServerConfig -> (c -> m ()) -> m ()
+runTransportServerWith run threadLabel serverParams cfg server = do
```
epoberezkin, line 71: I think it's better to also pass getSocket there, I don't think we have a case for custom "run"

PR #881, opened 2023-10-31
hybrid shared secret X25519 + sntrup761

epoberezkin:


src/Simplex/Messaging/Crypto/SNTRUP761/Bindings/RNG.hs
```diff
@@ -1,48 +1,31 @@
...
     copyByteArrayToPtr bs buf
-  pure RNG {rngContext = nullPtr, rngFunc}
-  where
-    swap (a, b) = (b, a)
-
-freeRNG :: RNG -> IO ()
-freeRNG RNG {rngFunc} = freeHaskellFunPtr rngFunc
 
-type RNGContext = Ptr RNG
+type RNGContext = Ptr Void
```
dpwiz, line 25: `type RNGContext = Ptr RNG` was a mistake in 

After removing the RNG wrapper, the type isn't used anywhere else. It's safe to put in into RNGFunc and FFI sigs directly. 

There are some options to pick, depending on your intention:

* `Ptr ()` to match C `void *`.
* `Ptr Void` if you want to forbid having anything but `nullPtr` (a pointer to a value that you can't create).
* `newtype RNGContext = RNGContext (Ptr RNGContext)` (circular, yes) - a convention for opaque pointers to be used in the C-land. E.g. `chachaCtx = RNGContext nullPtr` to instantiate. Newtypes over pointers are FFI-marshallable.
* Something fancy to tie RNG with its context?

dpwiz (approved):

PR #917, opened 2023-12-08
core: wrap message sending in batching monad

dpwiz:


src/Simplex/Messaging/Agent.hs
```diff
@@ -275,7 +278,7 @@ resubscribeConnections c = withAgentEnv c . resubscribeConnections' c
 
 -- | Send message to the connection (SEND command)
 sendMessage :: AgentErrorMonad m => AgentClient -> ConnId -> MsgFlags -> MsgBody -> m AgentMsgId
-sendMessage c = withAgentEnv c .:. sendMessage' c
+sendMessage c connId msgFlags msg = withAgentEnv c . withConnLock c connId "sendMessage" . execAgentBatch c $ \b -> sendMessageB c b connId msgFlags msg
```
epoberezkin, line 281: I think sendMessage' needs to stay with this implementation, it's used twice

src/Simplex/Messaging/Agent.hs
```diff
@@ -475,7 +478,7 @@ processCommand c (connId, APC e cmd) =
     ACPT invId ownCInfo -> (,OK) <$> acceptContact' c connId True invId ownCInfo SMSubscribe
     RJCT invId -> rejectContact' c connId invId $> (connId, OK)
     SUB -> subscribeConnection' c connId $> (connId, OK)
-    SEND msgFlags msgBody -> (connId,) . MID <$> sendMessage' c connId msgFlags msgBody
+    SEND msgFlags msgBody -> (connId,) . MID <$> withConnLock c connId "sendMessage" (execAgentBatch c $ \b -> sendMessageB c b connId msgFlags msgBody)
```
epoberezkin, line 481: this won't change in this case

src/Simplex/Messaging/Agent.hs
```diff
@@ -1053,25 +1059,25 @@ runCommandProcessing c@AgentClient {subQ} server_ = do
...
+    filter isActiveSndQ sqs
   pure msgId
 
-enqueueMessage :: forall m. AgentMonad m => AgentClient -> ConnData -> SndQueue -> MsgFlags -> AMessage -> m AgentMsgId
-enqueueMessage c cData@ConnData {connId} sq msgFlags aMessage = do
-  resumeMsgDelivery c cData sq
-  aVRange <- asks $ smpAgentVRange . config
-  msgId <- storeSentMsg $ maxVersion aVRange
-  queuePendingMsgs c sq [msgId]
+enqueueMessage :: AgentMonad' m => AgentClient -> AgentBatch m -> ConnData -> SndQueue -> MsgFlags -> AMessage -> BatchT AgentErrorType m AgentMsgId
```
epoberezkin, line 1071: same for this, if this new one is called enqueueMessageB, the old one would stay, and some calls won't change

PR #920, opened 2023-12-13
use ChaChaDRG as the source of randomness

epoberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -388,8 +388,8 @@ xftpDeleteSndFileRemote :: AgentErrorMonad m => AgentClient -> UserId -> SndFile
 xftpDeleteSndFileRemote c = withAgentEnv c .:. deleteSndFileRemote c
 
 -- | Create new remote host pairing
-rcNewHostPairing :: MonadIO m => m RCHostPairing
-rcNewHostPairing = liftIO newRCHostPairing
+rcNewHostPairing :: AgentErrorMonad m => AgentClient -> m RCHostPairing
+rcNewHostPairing c = withAgentEnv c $ liftIO . newRCHostPairing =<< asks random
```
dpwiz, line 392: A very roundabout way to fetch a field of a known type.

Perhaps something like `rcNewHostPairing = newRCHostPairing . random . agentEnv` ?

spaced4ndy (approved):

src/Simplex/Messaging/Agent.hs
epoberezkin: it just follows the pattern elsewhere...

PR #944, opened 2024-01-02
xftp-server: fix stats

dpwiz:


src/Simplex/FileTransfer/Server.hs
```diff
@@ -391,14 +397,20 @@ restoreServerStats = asks (serverStatsBackupFile . config) >>= mapM_ restoreStat
     restoreStats f = whenM (doesFileExist f) $ do
       logInfo $ "restoring server stats from file " <> T.pack f
       liftIO (strDecode <$> B.readFile f) >>= \case
-        Right d -> do
+        Right d@FileServerStatsData {_filesCount = statsFilesCount} -> do
           s <- asks serverStats
           fs <- readTVarIO . files =<< asks store
-          let _filesCount = length $ M.keys fs
-              _filesSize = M.foldl' (\n -> (n +) . fromIntegral . size . fileInfo) 0 fs
+          (_filesCount, _filesSize) <- foldM collect (0, 0) fs
```
epoberezkin, line 403: We already have usedStorage in store that is calculated when it's loaded - maybe we can avoid this fold?

epoberezkin: and I think the current count in stats is not the uploaded, but registered files, so it'll never match? so we could just compare count in stats with size of file store map, and size in stats with usedStorage in store.

PR #947, opened 2024-01-03
tests: add sntrup761 benchmarks

dpwiz:


dpwiz:
now it ab/benchmark-target

PR #954, opened 2024-01-09
SMP proxy: protocol

dpwiz:


rfcs/2023-09-12-second-relays.md
```diff
@@ -97,11 +97,11 @@ This would also reduce the difference in how the traffic looks to the observer -
...
+1. Client requests proxy to create session with the relay by sending `proxy` command with the SMP relay address and optional proxy basic AUTH (below). It should be possible to batch multiple session requests into one block, to reduce traffic.
 
-2. Proxy connects to SMP relay, negotiating a shared secret in the handshake that will be used to encrypt all sender blocks inside TLS (proxy-relay encryption). SMP relay also returns in handshake its temporary DH key to agree e2e encryption with the client (sender-relay encryption, to hide metadata sent to the destination relay from proxy).
+2. Proxy connects to SMP relay, and uses `p_handshake` command to negotiate a shared secret in the handshake that will be used to encrypt all sender blocks inside TLS (proxy-relay encryption). SMP relay also returns in handshake its temporary DH key to agree e2e encryption with the client (sender-relay encryption, to hide metadata sent to the destination relay from proxy).
```
epoberezkin, line 102: possibly, it could be in protocol handshake, there are pros/cons

epoberezkin: in this case key lifecycle will be longer than session lifecycle

notes-flow.txt
```diff
@@ -0,0 +1,23 @@
+common:
+  corrId - random BS, used as CbNonce
+  entityId - p2r tlsUniq
```
epoberezkin, line 3: don't think we have to use tlsuniq as session ID, can be just random? Or will it be included as sessionId in e2e encrypted commands to prevent replays?

rfcs/2023-09-12-second-relays.md
```diff
@@ -126,29 +126,29 @@ Sequence diagram for sending the message via SMP proxy:
...
-     |                            |                                 |                          |
-     |                            |     TLS(R:p2r(s2r(OK/ERR)))     |                          |
-     |     TLS(R:s2r(OK/ERR))     | <------------------------------ |                          |
-     | <------------------------- |                                 | TLS(MSG(r2c(e2e(msg))))  |
-     |                            |                                 | -----------------------> |
-     |                            |                                 |                          |
-     |                            |                                 |        TLS(ACK)          |
+     |                            |         `r_response` (p2r)      |                          |
+     |                            | <------------------------------ |                          |
+     |       `r_response`         |                                 |           `MSG`          |
```
epoberezkin, line 143:
```suggestion
     |       `r_response` (s2r)   |                                 |           `MSG`          |
```

src/Simplex/Messaging/Protocol.hs
```diff
@@ -300,6 +301,9 @@ data Command (p :: Party) where
   PING :: Command Sender
   -- SMP notification subscriber commands
   NSUB :: Command Notifier
+  PROXY :: SMPServer -> Maybe BasicAuth -> Command Sender -- request a relay server connection by URI
```
epoberezkin, line 304: there can be no auth here, it's only for creating queues

src/Simplex/Messaging/Protocol.hs
```diff
@@ -300,6 +301,9 @@ data Command (p :: Party) where
...
+  PHS :: Command Sender -- identify connection as a proxy, set up shared key
+  PFWD :: C.PublicKeyX25519 -> ByteString -> Command Sender -- use CorrId as CbNonce
```
epoberezkin, line 306:
```suggestion
  PXY :: SMPServer -> Command Sender -- request a relay server connection by URI
  PPXY :: C.PublicKeyX25519 -> Command Sender -- identify connection as a proxy, set up shared key
  PFWD :: C.PublicKeyX25519 -> ByteString -> Command Sender -- use CorrId as CbNonce
```

PPXY must have a key for the session encryption

src/Simplex/Messaging/Protocol.hs
```diff
@@ -336,6 +340,8 @@ data BrokerMsg where
   MSG :: RcvMessage -> BrokerMsg
   NID :: NotifierId -> RcvNtfPublicDhKey -> BrokerMsg
   NMSG :: C.CbNonce -> EncNMsgMeta -> BrokerMsg
+  RKEY :: C.PublicKeyX25519 -> C.Signature C.Ed25519 -> BrokerMsg -- TLS-signed server key for proxy shared secret and initial sender key
```
epoberezkin, line 343:
```suggestion
  RKEY :: C.PublicKeyX25519 -> Cert (?) -> C.Signature C.Ed25519 -> BrokerMsg -- TLS-signed server key for proxy shared secret and initial sender key. Sent by relay to proxy
  PRKEY :: C.PublicKeyX25519 -> Cert -> C.Signature C.Ed25519 -> BrokerMsg - sent by proxy to client.
```

I am not sure how this signature can be validated here, as we do not have the key, only fingerprint of the certificate, so possibly the entire certificate that the relay received in TLS handshake (two-cert self-signed chain) should be included in this command. Then we can validate the signature.

src/Simplex/Messaging/Protocol.hs
```diff
@@ -336,6 +340,8 @@ data BrokerMsg where
...
+  RRES :: ByteString -> BrokerMsg -- Encrypted response. E.g. relay to client, hidden from proxy
```
epoberezkin, line 344:
```suggestion
  RRES :: ByteString -> BrokerMsg -- Encrypted response. E.g. relay to client, hidden from proxy
  PRRES :: ByteString -> BrokerMsg -- Encrypted response. E.g. relay to client, hidden from proxy
```

I would use different newtypes for these bytestrings, with maybe max-length encoded in types, to make sure we correctly wrap/unwrap encryption layers.

src/Simplex/Messaging/Protocol.hs
epoberezkin: possibly type encoded lengths / newtypes instead of byte strings to wrap/unwrap correctly. The information about encryption layers should be encoded in types, or it's really easy to get confused here...

epoberezkin: or was BasicAuth here to allow usage of proxy?

rfcs/2023-09-12-second-relays.md
```diff
@@ -167,27 +167,33 @@ Below diagram shows the encrypttion layers for `forward` and `response` commands
 -----------------             -----------------  -- TLS --  -----------------             -----------------
 ```
 
-When proxy connects to SMP relay it would indicate in the handshake that it will use proxy protocol and the SMP relay would expect the same `forward` commands and reply with `response`s.
+When proxy connects to SMP relay it would indicate in the handshake that it will act as a proxy and the SMP relay would expect the same `forward` commands and reply with `response`s.
```
epoberezkin, line 170: Most likely s2r layer will be removed if entity ID and command encryption is added.

Correlation ID should be random, possible from a separate DRG per session, to minimise contention.

PR #956, opened 2024-01-10
smp: check for progress when resubscribing

dpwiz:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -572,7 +572,12 @@ reconnectSMPClient n tc c tSess@(_, srv, _) = do
         unless (null conns) $ notifySub "" $ UP srv conns
       let (tempErrs, finalErrs) = partition (temporaryAgentError . snd) errs
       liftIO $ mapM_ (\(connId, e) -> notifySub connId $ ERR e) finalErrs
-      mapM_ (throwError . snd) $ listToMaybe tempErrs
+      forM_ (listToMaybe tempErrs) $ \(_, err) ->
+        if allowClose && null okConns && S.null cs && null finalErrs
+          then do
+            liftIO $ closeClient c smpClients tSess
+            throwError $ BROKER (B.unpack $ strEncode srv) (RESPONSE "TODO: new error for no progress")
+          else throwError err
```
epoberezkin, line 580:
```suggestion
      forM_ (listToMaybe tempErrs) $ \(_, err) ->
        liftIO . when (allowClose && null okConns && S.null cs && null finalErrs) $
          closeClient c smpClients tSess
        throwError err
```

epoberezkin: I think we can just throw temp error as usual, as the log will show disconnect/connect event in this case, so we know it's working - no need for new error.

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -572,7 +575,11 @@ reconnectSMPClient n tc c tSess@(_, srv, _) = do
         unless (null conns) $ notifySub "" $ UP srv conns
       let (tempErrs, finalErrs) = partition (temporaryAgentError . snd) errs
       liftIO $ mapM_ (\(connId, e) -> notifySub connId $ ERR e) finalErrs
-      mapM_ (throwError . snd) $ listToMaybe tempErrs
+      forM_ (listToMaybe tempErrs) $ \(_, err) -> do
+        when (null okConns && S.null cs && null finalErrs) $
+          liftIO $
+            closeClient' c smpClients tSess
```
epoberezkin, line 581:
```suggestion
        when (null okConns && S.null cs && null finalErrs) . liftIO $
          closeClient' c smpClients tSess
```

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -535,33 +529,37 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
...
-        maxTC <- asks $ maxSubscriptionTimeouts . config
-        let err = if tc' >= maxTC then CRITICAL True else INTERNAL
-            msg = show tc' <> " consecutive subscription timeouts: " <> show (length qs) <> " queues, transport session: " <> show tSess
-        atomically $ writeTBQueue (subQ c) ("", "", APC SAEConn $ ERR $ err msg)
+      withRetryIntervalCount ri $ \_ _ loop ->
+        unlessM (reconnectSMPClient timeoutCounts c tSess) loop `catchAgentError` const loop
+
+reconnectSMPClient :: forall m. AgentMonad m => TVar Int -> AgentClient -> SMPTransportSession -> m Bool
+reconnectSMPClient tc c tSess@(_, srv, _) = do
+  qs <- getPending
```
epoberezkin, line 547:
```suggestion
  qs <- atomically (RQ.getSessQueues tSess $ pendingSubs c)
  if null qs
    then pure False
    else do
      ...
      pure True
```

or even move it outside, to tryReconnectSMPClient

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -535,35 +530,43 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
...
-        tc' <- atomically $ stateTVar tc $ \i -> (i + 1, i + 1)
-        maxTC <- asks $ maxSubscriptionTimeouts . config
-        let err = if tc' >= maxTC then CRITICAL True else INTERNAL
-            msg = show tc' <> " consecutive subscription timeouts: " <> show (length qs) <> " queues, transport session: " <> show tSess
-        atomically $ writeTBQueue (subQ c) ("", "", APC SAEConn $ ERR $ err msg)
+      withRetryIntervalCount ri $ \_ _ loop -> do
+        pending <- atomically $
+          RQ.getSessQueues tSess (pendingSubs c) >>= \case
+            [] -> Nothing <$ cleanup
+            q : qs -> pure $ Just (q :| qs)
```
epoberezkin, line 548:
```suggestion
        pending <- atomically $ do
          qs <- RQ.getSessQueues tSess (pendingSubs c) >>= \case
          when (null qs) cleanup
          pure qs
```

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -535,35 +530,43 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
...
+        forM_ pending $ \qs -> do
```
epoberezkin, line 549:
```suggestion
        pending <- atomically $ do
          qs <- RQ.getSessQueues tSess (pendingSubs c) >>= \case
          when (null qs) cleanup
          pure qs
        forM_ (L.nonEmpty pending) $ \qs -> do
```

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -535,35 +530,43 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
         notifySub connId cmd = atomically $ writeTBQueue (subQ c) ("", connId, APC (sAEntity @e) cmd)
 
 reconnectServer :: AgentMonad m => AgentClient -> SMPTransportSession -> m ()
-reconnectServer c tSess = newAsyncAction tryReconnectSMPClient $ reconnections c
+reconnectServer c tSess =
```
epoberezkin, line 533: all calls to reconnectServer should somehow be combined with adding pending queue, so that the check that there is no reconnect loop is also atomic

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -535,35 +530,43 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
...
+  atomically setup >>= \case
+    Left reserved -> do
+      a <- async $ tryReconnectSMPClient `E.catchAny` const (atomically cleanup)
+      atomically (takeTMVar reserved >> putTMVar reserved (Just a))
+    Right running -> atomically $ tryReadTMVar running >>= mapM_ (maybe retry waitSTM)
```
epoberezkin, line 538: this is very difficult to understand, needs to be much simpler

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -535,35 +530,43 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
...
-            msg = show tc' <> " consecutive subscription timeouts: " <> show (length qs) <> " queues, transport session: " <> show tSess
-        atomically $ writeTBQueue (subQ c) ("", "", APC SAEConn $ ERR $ err msg)
+      withRetryIntervalCount ri $ \_ _ loop -> do
+        pending <- atomically $
+          RQ.getSessQueues tSess (pendingSubs c) >>= \case
+            [] -> Nothing <$ cleanup
+            q : qs -> pure $ Just (q :| qs)
+        forM_ pending $ \qs -> do
+          reconnectSMPClient timeoutCounts c tSess qs `catchAgentError` \_ -> pure ()
+          loop
```
epoberezkin, line 551: I am still thinking about growing timeout here, maybe we can somehow end it sooner, not sure.

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -529,41 +532,49 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
...
-reconnectServer c tSess = newAsyncAction tryReconnectSMPClient $ reconnections c
+resubscribeSMPSession :: AgentMonad m => AgentClient -> SMPTransportSession -> m ()
+resubscribeSMPSession c@AgentClient {smpSubWorkers} tSess =
+  atomically (getTSessVar tSess smpSubWorkers) >>= either newSubWorker (\_ -> pure ())
   where
-    tryReconnectSMPClient aId = do
+    newSubWorker v = do
+      subWorkerId <- atomically $ stateTVar (workerSeq c) $ \next -> (next, next + 1)
+      subWorkerAsync <- async $ runSubWorker subWorkerId `E.catchAny` const (atomically $ cleanup subWorkerId)
+      atomically $ putTMVar v SubWorker {subWorkerId, subWorkerAsync}
```
spaced4ndy, line 547: is there something that prevents multiple workers for the same session with different IDs? Check for session and adding to map are in separate atomically. Or all but one are killed afterwards? I may be missing something..

spaced4ndy: killed in `cleanup`?

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -529,41 +532,49 @@ getSMPServerClient c@AgentClient {active, smpClients, msgQ} tSess@(userId, srv,
...
-reconnectServer :: AgentMonad m => AgentClient -> SMPTransportSession -> m ()
-reconnectServer c tSess = newAsyncAction tryReconnectSMPClient $ reconnections c
+resubscribeSMPSession :: AgentMonad m => AgentClient -> SMPTransportSession -> m ()
+resubscribeSMPSession c@AgentClient {smpSubWorkers} tSess =
+  atomically (getTSessVar tSess smpSubWorkers) >>= either newSubWorker (\_ -> pure ())
   where
-    tryReconnectSMPClient aId = do
+    newSubWorker v = do
+      subWorkerId <- atomically $ stateTVar (workerSeq c) $ \next -> (next, next + 1)
+      subWorkerAsync <- async $ runSubWorker subWorkerId `E.catchAny` const (atomically $ cleanup subWorkerId)
```
spaced4ndy, line 546: why not catchAgentError?

dpwiz: Doesn't matter, the exact error is ignored.

src/Simplex/Messaging/Agent/Client.hs
dpwiz: It is `getTSessVar` that does the reservation by putting empty TMVar at the key. `putTMVar` here only fills it it with the actual value.

spaced4ndy: right, thanks 🤦‍♂️

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: see the last commit, it's now clear from the types that there can be only IO exception

PR #957, opened 2024-01-12
agent: do not create user record in new databases

epoberezkin:


src/Simplex/Messaging/Agent/Store/SQLite/Migrations.hs
```diff
@@ -128,9 +128,13 @@ run st = \case
   where
     runUp Migration {name, up, down} = withTransaction' st $ \db -> do
       when (name == "m20220811_onion_hosts") $ updateServers db
-      insert db >> execSQL db up
+      insert db >> execSQL db up'
       where
         insert db = DB.execute db "INSERT INTO migrations (name, down, ts) VALUES (?,?,?)" . (name,down,) =<< getCurrentTime
+        -- we could add an optional field for new database migration to the list
```
spaced4ndy, line 134: I don't think it's a good idea to legitimize this approach, this is an exceptional case. Each such "revised" migration would split user base into separate categories in terms of possible errors
```suggestion
```

src/Simplex/Messaging/Agent/Store/SQLite/Migrations/M20230110_users.hs
```diff
@@ -27,3 +27,18 @@ UPDATE connections SET user_id = 1;
...
+
+new_m20230110_users :: Query
+new_m20230110_users =
+  [sql|
+CREATE TABLE users (
+  user_id INTEGER PRIMARY KEY AUTOINCREMENT
+);
+
+ALTER TABLE connections ADD COLUMN user_id INTEGER NOT NULL
+REFERENCES users ON DELETE CASCADE;
```
spaced4ndy, line 39: let's not at least make schema different..
```suggestion
ALTER TABLE connections ADD COLUMN user_id INTEGER CHECK (user_id NOT NULL)
REFERENCES users ON DELETE CASCADE;
```

spaced4ndy (approved):

PR #962, opened 2024-01-14
agent: fix race condition in subscription worker (results in stuck delivery), better error handling in batch operations

epoberezkin:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -417,7 +418,7 @@ newAgentClient InitialAgentServers {smp, ntf, xftp, netCfg} agentEnv = do
   smpSubWorkers <- TM.empty
   asyncClients <- newTAsyncs
   agentStats <- TM.empty
-  clientId <- stateTVar (clientCounter agentEnv) $ \i -> let i' = i + 1 in (i', i')
+  clientId <- maybe nextClientId pure clientId_
```
spaced4ndy, line 421: wouldn't it work in a way that nextClientId is untested? maybe it'd better be a separate label?

PR #967, opened 2024-01-16
agent: fix potential race when good client can be removed instead of bad for the same transport session

epoberezkin:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -673,18 +664,18 @@ newProtocolClient ::
...
   (AgentClient -> TransportSession msg -> m ()) ->
   ClientVar msg ->
   m (Client msg)
-newProtocolClient c tSess@(userId, srv, entityId_) clients connectClient clientConnected clientVar = tryConnectClient pure tryConnectAsync
+newProtocolClient c tSess@(userId, srv, entityId_) clients connectClient clientConnected v = tryConnectClient pure tryConnectAsync
   where
     tryConnectClient :: (Client msg -> m a) -> m () -> m a
     tryConnectClient successAction retryAction =
-      tryError connectClient >>= \r -> case r of
+      tryError (connectClient v) >>= \r -> case r of
```
dpwiz, line 675: `tryError` doesn't catch IO errors (it's only `liftIOEIther` inside).

This may leave session map in perpetual `key present, TMVar empty` state.
`waitForProtocol` don't remove the key and `newProtocol` only removeTSessVar on agent error here.

PR #982, opened 2024-02-04
smp: command authorization

epoberezkin:
Scope:
- supports crypto_box authenticators - planned to enable in SMPv7 for senders' commands only to provide sender repudiation on the server level (see RFC).
- removes tlsunique from sent bytes, but uses it during signing and verifying (to prevent replay attacks) - this is controlled by implySessId property of THandle, as it is in different SMP and NTF protocol versions.
- added batching to NTF protocol (no commands are batched, but until all clients migrate we cannot remove support for non-batched encoding.
- raised min SMP protocol to v4 (batching).
- correlation IDs are now random 24 bytes, to avoid future correlation by session by destination relays when commands will be sent via sending proxies. They are also used as nonces for authenticators.
- handshake headers are extended to include public key, but client does not use its session key for anything as per-queue keys are used. See encryption RFC. Sending proxies would require a separate encryption layer for transmission in any case, so adding general transmission encryption inside TLS will not help it - so it is out of scope of this PR.

Test required: new/existing connections via 16 combinations of old (pre v5.5.3)/new clients and old/new servers participating in the connection (4 parties) - 32 tests in total.

The same testing will need to be repeated once server versions are raised in code; old (pre v5.5.3) clients will not work with the new servers but new clients will continue supporting old servers (see migration plan in Deniability RFC).


rfcs/2024-02-03-deniability.md
```diff
@@ -0,0 +1,51 @@
...
+
+While it was never claimed that the messaging protocol provides deniability, the deniability is often mentioned as one of the important qualities of double ratchet algorithm used in the innermost layer of e2e encryption, so without explicit disclaimer of deniability limitations, it may be assumed by the users that the system as a whole provides the same level of deniability as the double ratchet algorithm, which currently is not the case.
+
+While societal understanding and legal acceptance of repudiation is arguable, there was less than a decade since this quality became widely available in Signal - legal systems take longer to evolve.  While the argument that the message was forged is likely to be accepted in the usual court cases with the ordinary people, it is likely to be considered in cases with high profile defendants, who can reasonably claim that they are the target of smear campaign and are being attributed something that they never sent - the statement that the message is forged is reasonable in such cases, and it provides plausible deniability, and the cryptographic experts invited to the hearing would attest to that.
```
spaced4ndy, line 11:
```suggestion
While societal understanding and legal acceptance of repudiation is arguable, there was less than a decade since this quality became widely available in Signal - legal systems take longer to evolve.  While the argument that the message was forged is unlikely to be accepted in the usual court cases with the ordinary people, it is likely to be considered in cases with high profile defendants, who can reasonably claim that they are the target of smear campaign and are being attributed something that they never sent - the statement that the message is forged is reasonable in such cases, and it provides plausible deniability, and the cryptographic experts invited to the hearing would attest to that.
```

spaced4ndy: probably meant that

src/Simplex/Messaging/Notifications/Transport.hs
```diff
@@ -1,72 +1,136 @@
...
+    authPubKey <- authEncryptCmdsP (maxVersion ntfVersionRange) $ C.getSignedExact <$> smpP
+    pure NtfServerHandshake {ntfVersionRange, sessionId, authPubKey}
+
+encodeAuthEncryptCmds :: Encoding a => Version -> Maybe a -> ByteString
+encodeAuthEncryptCmds v k
+  | v >= authBatchCmdsNTFVersion = maybe "" smpEncode k
+  | otherwise = ""
+
+authEncryptCmdsP :: Version -> Parser a -> Parser (Maybe a)
+authEncryptCmdsP v p = if v >= authBatchCmdsNTFVersion then Just <$> p else pure Nothing
```
spaced4ndy, line 75: maybe version should be named differently? seems like it's not only adds batching?

PR #985, opened 2024-02-07
transport: fetch and store server certificate

dpwiz:


src/Simplex/Messaging/Crypto.hs
```diff
@@ -1384,3 +1405,9 @@ keyError :: (a, [ASN1]) -> Either String b
 keyError = \case
   (_, []) -> Left "unknown key algorithm"
   _ -> Left "more than one key"
+
+x509ToPrivateSign :: PrivKey -> Either String APrivateSignKey
```
epoberezkin, line 1409: this seems unnecessary, there is already x509ToPrivate, it can be combined with type check

epoberezkin: with CryptoPrivateKey instance

src/Simplex/Messaging/Transport.hs
```diff
@@ -194,13 +201,25 @@ data TProxy c = TProxy
 
 data ATransport = forall c. Transport c => ATransport (TProxy c)
 
+getServerVerifyKey :: Transport c => c -> Either String C.APublicVerifyKey
+getServerVerifyKey c =
+  case getServerCerts c of
+    X.CertificateChain (server : _ca) -> fromX509 . X.certPubKey . X.signedObject $ X.getSigned server
+    _ -> Left "no certificate chain"
+  where
+    fromX509 = \case
```
epoberezkin, line 210: this is also like x509ToPublic with type check using CryptoPublicKey

src/Simplex/Messaging/Transport.hs
```diff
@@ -325,30 +348,45 @@ data ClientHandshake = ClientHandshake
...
 
 instance Encoding ServerHandshake where
-  smpEncode ServerHandshake {smpVersionRange, sessionId, authPubKey} =
-    smpEncode (smpVersionRange, sessionId) <> encodeAuthPubKey (maxVersion smpVersionRange) authPubKey
+  smpEncode ServerHandshake {smpVersionRange, sessionId, authPubKey, certChain} =
+    B.concat
+      [ smpEncode (smpVersionRange, sessionId),
+        encodeAuthEncryptCmds maxV $ encodeChain <$> certChain,
+        encodeAuthEncryptCmds maxV $ C.SignedObject <$> authPubKey
+      ]
```
epoberezkin, line 364: you cannot make it independently of the version, it won't work with the old clients who will only be able to parse smpVersionRange and sessionId

epoberezkin: ah, this checks version twice... this seems unnecessarily confusing, needs to change somehow

dpwiz: I thought maybe we will add chain and pubkey in different versions. Better put them under one check ofc, if possible.

PR #990, opened 2024-02-14
remove support for old versions

epoberezkin:


src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -543,11 +542,11 @@ createConn_ gVar cData create = checkConstraint SEConnDuplicate $ case cData of
   ConnData {connId} -> Right . (connId,) <$> create connId
 
 createNewConn :: DB.Connection -> TVar ChaChaDRG -> ConnData -> SConnectionMode c -> IO (Either StoreError ConnId)
-createNewConn db gVar cData@ConnData {userId, connAgentVersion, enableNtfs, duplexHandshake} cMode = do
+createNewConn db gVar cData@ConnData {userId, connAgentVersion, enableNtfs} cMode = do
   fst <$$> createConn_ gVar cData create
   where
     create connId =
-      DB.execute db "INSERT INTO connections (user_id, conn_id, conn_mode, smp_agent_version, enable_ntfs, duplex_handshake) VALUES (?,?,?,?,?,?)" (userId, connId, cMode, connAgentVersion, enableNtfs, duplexHandshake)
+      DB.execute db "INSERT INTO connections (user_id, conn_id, conn_mode, smp_agent_version, enable_ntfs, duplex_handshake) VALUES (?,?,?,?,?,?)" (userId, connId, cMode, connAgentVersion, enableNtfs, 1 :: Int)
```
dpwiz, line 549: `True`?

src/Simplex/Messaging/Server.hs
```diff
@@ -856,17 +856,14 @@ client clnt@Client {thVersion, subscriptions, ntfSubscriptions, rcvQ, sndQ, sess
...
+          Message {msgFlags, msgBody} -> RcvMsgBody {msgTs = msgTs', msgFlags, msgBody}
+          MessageQuota {} -> RcvMsgQuota msgTs'
           where
-            encrypt :: KnownNat i => MsgFlags -> C.MaxLenBS i -> RcvMessage
-            encrypt msgFlags body =
+            encrypt :: KnownNat i => C.MaxLenBS i -> RcvMessage
+            encrypt body =
               let encBody = EncRcvMsgBody $ C.cbEncryptMaxLenBS (rcvDhSecret qr) (C.cbNonce msgId') body
-               in RcvMessage msgId' msgTs' msgFlags encBody
+               in RcvMessage msgId' encBody
```
dpwiz, line 866: the binding is redundant now
`RcvMessage msgId' . EncRcvMsgBody $ ...`

tests/AgentTests/FunctionalAPITests.hs
```diff
@@ -387,20 +369,16 @@ testMatrix2 t runTest = do
...
 testRatchetMatrix2 :: ATransport -> (AgentClient -> AgentClient -> AgentMsgId -> IO ()) -> Spec
 testRatchetMatrix2 t runTest = do
   it "ratchet current" $ withSmpServer t $ runTestCfg2 agentCfg agentCfg 3 runTest
-  it "ratchet prev" $ withSmpServer t $ runTestCfg2 agentCfgRatchetVPrev agentCfgRatchetVPrev 3 runTest
-  it "ratchets prev to current" $ withSmpServer t $ runTestCfg2 agentCfgRatchetVPrev agentCfg 3 runTest
-  it "ratchets current to prev" $ withSmpServer t $ runTestCfg2 agentCfg agentCfgRatchetVPrev 3 runTest
-  it "ratchet v1" $ withSmpServer t $ runTestCfg2 agentCfgRatchetV1 agentCfgRatchetV1 3 runTest
-  it "ratchets v1 to current" $ withSmpServer t $ runTestCfg2 agentCfgRatchetV1 agentCfg 3 runTest
-  it "ratchets current to v1" $ withSmpServer t $ runTestCfg2 agentCfg agentCfgRatchetV1 3 runTest
+  -- Previous version for e2e is v1 - no longer supported.
```
dpwiz, line 377: can be self-enabling:

```
pendingV = if minVersion currentRange == maxVersion currentRange then pendingWith "same version" else id
```

src/Simplex/Messaging/Agent.hs
```diff
@@ -1236,14 +1224,8 @@ runSmpQueueMsgDelivery c@AgentClient {subQ} cData@ConnData {userId, connId, dupl
...
-                    -- Party joining connection sends REPLY after HELLO in v1,
-                    -- it is an error to send REPLY in duplexHandshake mode (v2),
-                    -- and this branch should never be reached as receive is created before the confirmation,
-                    -- so the condition is not necessary here, strictly speaking.
-                    _ -> unless (duplexHandshake == Just True) $ do
-                      srv <- getSMPServer c userId
-                      qInfo <- createReplyQueue c cData sq SMSubscribe srv
-                      void . enqueueMessage c cData sq SMP.noMsgFlags $ REPLY [qInfo]
+                    -- this branch should never be reached as receive queue is created before the confirmation,
+                    _ -> pure ()
```
dpwiz, line 1225: logError ?

src/Simplex/Messaging/Protocol.hs
```diff
@@ -189,6 +189,13 @@ import Simplex.Messaging.Transport.Client (TransportHost, TransportHosts (..))
...
 
+-- SMP client protocol version history:
+-- 1 - binary protocol encoding (1/1/2022)
+-- 2 - multiple server hostnames and versioned queue addresses (8/12/2022)
+
+srvHostnamesSMPClientVersion :: Version
+srvHostnamesSMPClientVersion = 2
+
 currentSMPClientVersion :: Version
 currentSMPClientVersion = 2
```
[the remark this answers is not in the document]
epoberezkin: easier to have number here I think

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -1126,17 +1135,22 @@ instance StrEncoding AConnectionRequestUri where
...
         pure . ACR SCMInvitation $ CRInvitationUri crData crE2eParams
-      CMContact -> pure . ACR SCMContact $ CRContactUri crData
+      -- contact links are adjusted to the minimum version supported by the agent
+      -- to preserve compatibility with the old links published online
+      CMContact -> pure . ACR SCMContact $ CRContactUri crData {crAgentVRange = adjustAgentVRange aVRange}
     where
       crModeP = "invitation" $> CMInvitation <|> "contact" $> CMContact
+      adjustAgentVRange vr =
+        let v = max duplexHandshakeSMPAgentVersion $ minVersion vr
+         in fromMaybe vr $ safeVersionRange v (max v $ maxVersion vr)
```
spaced4ndy, line 1153: seems strange to do it in parsing

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -967,7 +982,6 @@ data AMsgType
 instance Encoding AMsgType where
   smpEncode = \case
     HELLO_ -> "H"
-    REPLY_ -> "R"
```
spaced4ndy, line 970: keep commented for some time to not use in near future?

spaced4ndy: though we usually check ability to use new features via version, meaning if it's sent other side should also have it deleted.. so probably it's ok

src/Simplex/Messaging/Server/MsgStore.hs
```diff
@@ -3,14 +3,11 @@
...
-data MsgLogRecord = MLRv3 RecipientId Message | MLRv1 RecipientId RcvMessage
+data MsgLogRecord = MLRv3 RecipientId Message
 
 instance StrEncoding MsgLogRecord where
-  strEncode = \case
-    MLRv3 rId msg -> strEncode (Str "v3", rId, msg)
-    MLRv1 rId msg -> strEncode (rId, msg)
-  strP = "v3 " *> (MLRv3 <$> strP_ <*> strP) <|> MLRv1 <$> strP_ <*> strP
+  strEncode (MLRv3 rId msg) = strEncode (Str "v3", rId, msg)
+  strP = "v3 " *> (MLRv3 <$> strP_ <*> strP)
```
spaced4ndy, line 13: not sure about removing server encoding.. servers were more likely to stay not updated than clients? If server is updated now from old version it would lose queues. At least does store log parser sustain failure to parse some lines and continue to restore others?

src/Simplex/Messaging/Agent/Protocol.hs
epoberezkin: this is only done for contact addresses as they are more likely to be published long time ago, so they would be assume to have at least the minimal supported version. Would you put it in JOIN? The problem is that there are multiple entry points there... (sync/async)

src/Simplex/Messaging/Server/MsgStore.hs
epoberezkin: This is only for messages, not for queues. And yes, they would be ignored and logged as errors.

src/Simplex/Messaging/Agent/Protocol.hs
spaced4ndy: yes, probably would be more appropriate there.. I understand not wanting to duplicate logic across entry points, ok to keep as is.

PR #1015, opened 2024-02-26
agent: delay connection deletion to finish delivery of pending messages

spaced4ndy:


src/Simplex/Messaging/Agent.hs
```diff
@@ -1862,6 +1871,7 @@ cleanupManager c@AgentClient {subQ} = do
     deleteConns =
       withLock (deleteLock c) "cleanupManager" $ do
         void $ withStore' c getDeletedConnIds >>= deleteDeletedConns c
+        void $ withStore' c getDeletedWaitingDeliveryConnIds >>= deleteDeletedWaitingDeliveryConns c
```
spaced4ndy, line 1874: perhaps better to refactor deleteConnections_ so that connections are read outside of prepareDeleteConnections_ and passed to it, and call only deleteConnQueues here so that it can notify with DEL_CONN (or create a different function?)

src/Simplex/Messaging/Agent.hs
```diff
@@ -1452,39 +1452,49 @@ disableConn c connId = do
...
       rqs = concat $ M.elems rcvQs
       connIds' = M.keys rcvQs
   forM_ connIds' $ disableConn c
-  withStore' c $ forM_ (M.keys delRs) . deleteConn
+  -- ! delRs is not used to notify about the result in any of the calling functions,
+  -- ! it is only used to check results count in deleteConnections_;
+  -- ! if it was used to notify about the result, it might be necessary to differentiate
+  -- ! between completed deletions of connections, and deletions delayed due to wait for delivery (see deleteConn)
+  rs' <- catMaybes . rights <$> withStoreBatch' c (\db -> map (deleteConn db waitDelivery) (M.keys delRs))
+  forM_ rs' $ \cId -> notify ("", cId, APC SAEConn DEL_CONN)
```
spaced4ndy, line 1484: somewhy we didn't notify DEL_CONN here (perhaps because here only connections without rcv queues were deleted which is rare / not necessary to notify in this case / wasn't caught in tests), though it doesn't seem to contradict anything.

PR #1025, opened 2024-03-03
double ratchet versioning for post-quantum encryption

epoberezkin:


rfcs/2024-03-03-pqdr-version.md
```diff
@@ -0,0 +1,69 @@
...
+  toJSON (RatchetVR v1 v2) = toJSON (v1, v2)
+
+instance FromJSON RatchetVR where
+  parseJSON v = do
+    -- this also verifies that v2 > v1 (although we could remove JSON instances for VersionRange)
+    VersionRange v1 v2 <- parseJSON v
+    pure $ RatchetVR v1 v2
+```
+
+For connections, we could also make version used for the purposes of encoding dependent on the PQ being enabled, and version for decoding taken from message header, but then we'd have to not only upgrade ratchets but the connection as well every time PQ mode changes.
```
spaced4ndy, line 51: regarding connection version upgrade - there are two identical functions in store:
- `setConnAgentVersion`
- `setConnectionVersion`
probably doesn't affect proposition, just fyi in case you missed one

spaced4ndy: seems they became identical in a recent change, before one was also setting handshake field
https://github.com/simplex-chat/simplexmq/pull/990/files#diff-5830622f61972c8743b93cd2365db5108b7e838ae919e3c8fb368f7722d3014aR872

rfcs/2024-03-03-pqdr-version.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Solution
+
+Currently ratchet state stores version range. It's unclear what was the intended semantics of that version range - it simply stores the offered/supported version range at the time ratchet was initialised, but only a high bound is used to send in message headers, and it is never upgraded. In JSON this range is encoded as tuple (an array of two elements in JSON).
+
+We could continue using this range with the meaning of the lower bound to be "currently used ratchet version" and the meaning of higher boundary to be "maximum supported ratchet version". We could also use the version communicated in message headers to upgrade ratchet version, with the condition that upgrade should only happen if both sides want it. Currently it's defined by pqEnableKEM property in ratchet state. We could also make it more explicit by defining maximum version to which ratchet should upgrade. Given that irreversible upgrades are not very common, it is probably ok to keep it implicit.
```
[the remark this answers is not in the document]
epoberezkin: the semantics would remain in handshake, this change is only about versions in the ratchet, and the current version can only be higher than minVersion - this would prevent downgrades of the existing ratchets and will allow ratchet encodings depend on this current version.

rfcs/2024-03-03-pqdr-version.md
epoberezkin: so should we merge them into one?

rfcs/2024-03-03-pqdr-version.md
```diff
@@ -0,0 +1,69 @@
...
+- ability to opt-in into this functionality via "experimental" feature toggle, that enables post-quantum encryption in connections when both contacts enable this toggle.
+
+To have ability to downgrade the clients we have two options:
+- roll-out this functionality in two stages: 1) roll-out clients support but do not enable the new version, and then 2) upgrade client version. The problem here is that the clients won't be able to opt-in into this experiment.
+- make offered range dependent on experimental feature being enabled. Currently we have an option to enable PQ encryption in agent API, and this option can be used as a proxy to maxium supported protocol version - if the option is passed, it can be seen as an indication that higher version range (or version) should offered (or accepted).
+
+## Solution
+
+Currently ratchet state stores version range. It's unclear what was the intended semantics of that version range - it simply stores the offered/supported version range at the time ratchet was initialised, but only a high bound is used to send in message headers, and it is never upgraded. In JSON this range is encoded as tuple (an array of two elements in JSON).
```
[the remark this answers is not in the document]
epoberezkin: I don't think this coupling is needed - ratchet version would upgrade independently and it won't generate any keys otherwise.

dpwiz (approved):

PR #1065, opened 2024-03-23
replace base64-bytestring with base64

dpwiz:


src/Simplex/Messaging/Agent/Store/SQLite.hs
```diff
@@ -2248,7 +2249,7 @@ createWithRandomId' gVar create = tryCreate 3
           | otherwise -> pure . Left . SEInternal $ bshow e
 
 randomId :: TVar ChaChaDRG -> Int -> IO ByteString
-randomId gVar n = atomically $ U.encode <$> C.randomBytes n gVar
+randomId gVar n = atomically $ extractBase64 . U.encodeBase64' <$> C.randomBytes n gVar
```
dpwiz, line 2252: url format allows being unpadded for any size, but it breaks entity-encoding expectations elsewhere.

PR #1070, opened 2024-03-25
add key to XFTP server control port command

epoberezkin:


src/Simplex/FileTransfer/Server.hs
```diff
@@ -207,13 +207,15 @@ xftpServer cfg@XFTPServerConfig {xftpPort, transportConfig, inactiveClientExpira
                 Left err -> hPutStrLn h ("error: " <> err) >> cpLoop h
             processCP h = \case
               CPStatsRTS -> E.tryAny getRTSStats >>= either (hPrint h) (hPrint h)
-              CPDelete fileId -> unliftIO u $ do
+              CPDelete fileId fKey -> unliftIO u $ do
```
spaced4ndy, line 210: why is it needed to check key if control port can only be used by server admin? are IDs reused making race on deletion possible?

epoberezkin: to reduce possibility of attack - fileId is used as file name.

PR #1102, opened 2024-04-17
xftp: check proxy before downloading from unknown server

spaced4ndy:


src/Simplex/FileTransfer/Agent.hs
```diff
@@ -214,6 +217,19 @@ runXFTPRcvWorker c srv Worker {doWork} = do
...
+          knownSrvs_ <- atomically $ TM.lookup userId xftpServers
+          let srvKnown = maybe False (\knownSrvs -> srv `elem` L.map protoServer knownSrvs) knownSrvs_
+          -- assuming specific XFTP client will have the same socksProxy config
+          (_, NetworkConfig {socksProxy, hostMode, requiredHostMode}) <- readTVarIO useNetworkConfig
+          let usingProxy =
+                -- on Android we have direct information about using proxy
+                isJust socksProxy
+                  -- on iOS we can infer based on hostMode and requiredHostMode
+                  || ((hostMode == HMOnionViaSocks || hostMode == HMOnion) && requiredHostMode)
+          pure $ not srvKnown && not usingProxy
```
spaced4ndy, line 232: it may make sense to separate this into new agent api to not repeat logic in chat and UI - similar to `connectionPlan` - `checkUnknownNoProxy`

epoberezkin: It is better to decompose it differently.

1) whether the file has approved relays - this check should be done in the chat (and can be exposed via the API), at the time that the decision to download is made in the UI. That sets the parameter on the file whether it is approved relays (either because user approved them or because they are in approved list).
2) is current connection via Tor - this would happen both in UI and in the agent. 

This way the change in known servers won't affect the outcome at the time of the check - e.g. user removing known proxies from the list shouldn't prevent completion of the download.

The agent would then only stop download if file relays are not approved.

Also, the term we should use is socksProxy, not just proxy, or it'll be confused with protocol proxies.

src/Simplex/FileTransfer/Transport.hs
```diff
@@ -214,6 +214,8 @@ data XFTPErrorType
     TIMEOUT
   | -- | bad redirect data
     REDIRECT {redirectError :: String}
+  | -- | cannot proceed with download from unknown server without proxy
+    UNKNOWN_NO_PROXY
```
epoberezkin, line 218: this is the error type that is sent in protocol, it shouldn't be here

src/Simplex/Messaging/Agent/Store/SQLite/Migrations/M20240417_rcv_files_only_via_proxy.hs
```diff
@@ -0,0 +1,18 @@
...
+
+module Simplex.Messaging.Agent.Store.SQLite.Migrations.M20240417_rcv_files_only_via_proxy where
+
+import Database.SQLite.Simple (Query)
+import Database.SQLite.Simple.QQ (sql)
+
+m20240417_rcv_files_only_via_proxy :: Query
+m20240417_rcv_files_only_via_proxy =
+    [sql|
+ALTER TABLE rcv_files ADD COLUMN only_via_proxy INTEGER NOT NULL DEFAULT 0;
```
epoberezkin, line 11:
```suggestion
ALTER TABLE rcv_files ADD COLUMN approved_relays INTEGER NOT NULL DEFAULT 0;
```

epoberezkin: so the agent wouldn't need to know whether the relays were approved by the user or because they were in the list of known relays at the time download started.

src/Simplex/FileTransfer/Agent.hs
```diff
@@ -112,8 +114,8 @@ closeXFTPAgent a = do
   where
     stopWorkers workers = atomically (swapTVar workers M.empty) >>= mapM_ (liftIO . cancelWorker)
 
-xftpReceiveFile' :: AgentClient -> UserId -> ValidFileDescription 'FRecipient -> Maybe CryptoFileArgs -> AM RcvFileId
-xftpReceiveFile' c userId (ValidFileDescription fd@FileDescription {chunks, redirect}) cfArgs = do
+xftpReceiveFile' :: AgentClient -> UserId -> ValidFileDescription 'FRecipient -> Maybe CryptoFileArgs -> Bool -> AM RcvFileId
+xftpReceiveFile' c userId (ValidFileDescription fd@FileDescription {chunks, redirect}) cfArgs onlyViaProxy = do
```
epoberezkin, line 118:
```suggestion
xftpReceiveFile' c userId (ValidFileDescription fd@FileDescription {chunks, redirect}) cfArgs approvedRelays = do
```

src/Simplex/FileTransfer/Agent.hs
```diff
@@ -214,6 +217,19 @@ runXFTPRcvWorker c srv Worker {doWork} = do
       when complete . lift . void $
         getXFTPRcvWorker True c Nothing
       where
+        unknownNoProxy :: AM Bool
+        unknownNoProxy = do
+          let AgentClient {xftpServers, useNetworkConfig} = c
+          knownSrvs_ <- atomically $ TM.lookup userId xftpServers
+          let srvKnown = maybe False (\knownSrvs -> srv `elem` L.map protoServer knownSrvs) knownSrvs_
+          -- assuming specific XFTP client will have the same socksProxy config
+          (_, NetworkConfig {socksProxy, hostMode, requiredHostMode}) <- readTVarIO useNetworkConfig
```
epoberezkin, line 226: you shouldn't use useNetworkConfig, there is getNetworkConfig

src/Simplex/FileTransfer/Agent.hs
```diff
@@ -192,8 +194,9 @@ runXFTPRcvWorker c srv Worker {doWork} = do
               atomically $ assertAgentForeground c
               loop
             retryDone e = rcvWorkerInternalError c rcvFileId rcvFileEntityId (Just fileTmpPath) (show e)
-    downloadFileChunk :: RcvFileChunk -> RcvFileChunkReplica -> AM ()
-    downloadFileChunk RcvFileChunk {userId, rcvFileId, rcvFileEntityId, rcvChunkId, chunkNo, chunkSize, digest, fileTmpPath} replica = do
+    downloadFileChunk :: RcvFileChunk -> RcvFileChunkReplica -> Bool -> AM ()
+    downloadFileChunk RcvFileChunk {userId, rcvFileId, rcvFileEntityId, rcvChunkId, chunkNo, chunkSize, digest, fileTmpPath} replica onlyViaProxy = do
+      when onlyViaProxy $ whenM unknownNoProxy $ throwError $ XFTP XFTP.UNKNOWN_NO_PROXY
```
epoberezkin, line 199:
```suggestion
      unlessM ((approvedRelays ||) <$> usesSocksProxy) $ throwError $ XFTP XFTP.UNKNOWN_NO_PROXY
```

also different error

epoberezkin: also this error should be AgentErrorType, not XFTPErrorType, so there may so other way to exit this function other than throwing error, so it's in AM monad.

src/Simplex/FileTransfer/Transport.hs
epoberezkin: should be in AgentErrorType, also the name is wrong and too long.

epoberezkin: AgentErrorType should have FILE constructor and all errors that are not sent via the protocol should be moved here. XFTP constructor in AgentErrorType should only be used for errors sent by servers to clients (and currently these errors are forwarded as INTERNAL, so it also needs to be fixed).

PR #1103, opened 2024-04-17
put smp errors into proxy wrappers

dpwiz:


src/Simplex/Messaging/Encoding/String.hs
```diff
@@ -75,9 +76,10 @@ instance StrEncoding Str where
   strEncode = unStr
   strP = Str <$> A.takeTill (== ' ') <* optional A.space
 
+-- inherited from ByteString, the parser only allows non-empty strings
 instance StrEncoding String where
-  strEncode = strEncode . B.pack
-  strP = B.unpack <$> strP
+  strEncode = strEncode . encodeUtf8 . T.pack
```
epoberezkin, line 81: it was on purpose really, I don't think it's a good idea...

PR #1106, opened 2024-04-19
proxy: agent implementation

epoberezkin:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -903,22 +1003,46 @@ withLogClient_ c tSess@(_, srv, _) entId cmdStr action = do
...
+    shouldUseProxy = do
+      cfg <- getNetworkConfig c
+      case smpProxyMode cfg of
+        SPMAlways -> pure True
+        SPMUnknown -> unknownServer
+        SPMUnprotected
+          | ipAddressProtected cfg destSrv -> pure False
+          | otherwise -> unknownServer
+        SPMNever -> pure False
+    unknownServer = maybe True (all ((destSrv /=) . protoServer)) <$> TM.lookup userId (userServers c)
```
spaced4ndy, line 1032: if I'm not mistaken this would not include disabled servers - agent is only passed active ("Use for new connections") servers. So it was ok in chat but it has a different meaning in agent

spaced4ndy: also why not use notElem, it likely would be way more straightforward to read

src/Simplex/Messaging/Client.hs
```diff
@@ -650,14 +669,14 @@ deleteSMPQueues = okSMPCommands DEL
 
 -- send PRXY :: SMPServer -> Maybe BasicAuth -> Command Sender
 -- receives PKEY :: SessionId -> X.CertificateChain -> X.SignedExact X.PubKey -> BrokerMsg
-createSMPProxySession :: SMPClient -> SMPServer -> Maybe BasicAuth -> ExceptT SMPClientError IO (SessionId, VersionSMP, C.PublicKeyX25519)
-createSMPProxySession c relayServ@ProtocolServer {keyHash = C.KeyHash kh} proxyAuth =
+connectSMPProxiedRelay :: SMPClient -> SMPServer -> Maybe BasicAuth -> ExceptT SMPClientError IO ProxiedRelay
+connectSMPProxiedRelay c relayServ@ProtocolServer {keyHash = C.KeyHash kh} proxyAuth =
   sendSMPCommand c Nothing "" (PRXY relayServ proxyAuth) >>= \case
-    -- XXX: rfc says sessionId should be in the entityId of response
+    -- TODO: rfc says sessionId should be in the entityId of response
```
spaced4ndy, line 675: just highlighting in case it got missed

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: because there is a function before comparison, and to use notElem would require using map

epoberezkin: yes, disabled will be treated as unknown

src/Simplex/Messaging/Client.hs
epoberezkin: for consistency with NEW it's not needed...

PR #1118, opened 2024-04-30
transport: reduce ping traffic

dpwiz:


src/Simplex/Messaging/Client.hs
```diff
@@ -504,15 +531,17 @@ createSMPQueue c (rKey, rpKey) dhKey auth subMode =
 --
 -- https://github.com/simplex-chat/simplexmq/blob/master/protocol/simplex-messaging.md#subscribe-to-queue
 subscribeSMPQueue :: SMPClient -> RcvPrivateAuthKey -> RecipientId -> ExceptT SMPClientError IO ()
-subscribeSMPQueue c rpKey rId =
+subscribeSMPQueue c@ProtocolClient {client_ = PClient {sendPings}} rpKey rId =
   sendSMPCommand c (Just rpKey) rId SUB >>= \case
-    OK -> return ()
+    OK -> liftIO . atomically $ writeTVar sendPings True
```
epoberezkin, line 536: why don't we enable it in case of MSG?

src/Simplex/Messaging/Client.hs
```diff
@@ -389,25 +396,45 @@ getProtocolClient g transportSession@(_, srv, _) cfg@ProtocolClientConfig {qSize
...
+              logDebug $ thost <> " : last response was " <> tshow diff <> " ago, delaying ping for " <> tshow (secondsToNominalDiffTime $ fromIntegral remaining / 1000000)
+              loop remaining
+            else do
+              logDebug $ thost <> " : last response was " <> tshow diff <> " ago, sending ping"
+              -- sendProtocolCommand/getResponse updates pingErrorCount
+              runExceptT (sendProtocolCommand c Nothing "" $ protocolPing @v @err @msg) >>= \case
+                Left PCEResponseTimeout -> do
+                  cnt <- readTVarIO pingErrorCount
+                  -- drop client when maxCnt of commands (pings or otherwise) have timed out in sequence, but only after some time has passed after last received response
+                  when (maxCnt == 0 || cnt < maxCnt || diff < recoverWindow) $ loop smpPingInterval
```
epoberezkin, line 433: the problem with this implementation is that if PING is never sent and the client keeps trying to send messages and they all timeout, the client will never be dropped.

epoberezkin: most likely the check to exit is needed in the first branch too

src/Simplex/Messaging/Client.hs
```diff
@@ -337,6 +342,7 @@ getProtocolClient g transportSession@(_, srv, _) cfg@ProtocolClientConfig {qSize
     mkProtocolClient :: TransportHost -> STM (PClient v err msg)
     mkProtocolClient transportHost = do
       connected <- newTVar False
+      sendPings <- newTVar False
       pingErrorCount <- newTVar 0
```
epoberezkin, line 346: probably better to rename to timeoutErrorCount

PR #1122, opened 2024-05-02
client: prevent sub actions from zombie sessions

dpwiz:


src/Simplex/Messaging/Agent.hs
```diff
@@ -1957,11 +1958,13 @@ getSMPServer c userId = withUserServers c userId pickServer
...
 subscriber :: AgentClient -> AM' ()
-subscriber c@AgentClient {msgQ} = forever $ do
+subscriber c@AgentClient {subQ, msgQ} = forever $ do
   t <- atomically $ readTBQueue msgQ
   agentOperationBracket c AORcvNetwork waitUntilActive $
     runExceptT (processSMPTransmission c t) >>= \case
-      Left e -> liftIO $ print e
+      Left e -> do
+        logError $ tshow e
+        atomically $ writeTBQueue subQ ("", "", APC SAENone $ AERR e)
```
dpwiz, line 1967: ERR is SAEConn message, but there's no connection entity to handle subscriber crashes.

Even if the processing sends the regular ERR inside, stuff thrown by ExceptT is ignored. Perhaps this should be in a separate pair of PRs.

PR #1132, opened 2024-05-06
proxy: fallback to direct connection if destination relay does not support proxy protocol

epoberezkin:


src/Simplex/Messaging/Client.hs
```diff
@@ -261,6 +270,7 @@ defaultNetworkConfig =
       requiredHostMode = False,
       sessionMode = TSMUser,
       smpProxyMode = SPMNever,
+      smpProxyFallback = SPFAllow,
```
spaced4ndy, line 273: maybe default should be SPFAllowProtected

epoberezkin: it will be that later, if we use it now - it will break everybody.

The idea of the initial opt-in is to test its working rather than providing "protection" - it will have to be explained in a blog post.

So the recommended option to test would be SPMUknown or SPMAlways (but with Tor, if they want to still protect IP addresses) and with SPFAllow, so it would test relays via proxy with/without tor, but without breaking anything - as long as we show in the message how it was delivered, it's cool.

A reasonable default for the future release would be SPMUnprotected + SPFAllowProtected - in which case it will use proxy only if Tor is not used, but would fail if destination server does not support proxy protocol.

PR #1161, opened 2024-05-20
core: use catMaybes to collect optional replies

dpwiz:


src/Simplex/Messaging/Agent.hs
```diff
@@ -2080,43 +2080,45 @@ data ACKd = ACKd | ACKPending
...
+          Right SMP.OK -> processSubOk rq
           Right msg@SMP.MSG {} -> do
-            processSubOk rq upConnIds
-            processSMP rq conn (toConnData conn) msg
-          Right r -> processSubErr rq $ unexpectedResponse r
-          Left e -> unless (temporaryClientError e) $ processSubErr rq e -- timeout/network was already reported
-        _ -> pure ()
-    STResponse {} -> pure () -- TODO process expired responses to sent messages
+            cId_ <- processSubOk rq
+            cId_ <$ processSMP rq conn (toConnData conn) msg
```
epoberezkin, line 2096: the problem here is that if processSMP throws exception (ExceptT or IO) then it won't return cId_ - it will be lost in the handler

PR #1172, opened 2024-05-25
core: send UP from initial subscriptions

dpwiz:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -1358,23 +1350,27 @@ subscribeQueues c qs = do
...
+          when (any isNothing subResults) resubscribe
+          let up = catMaybes $ L.toList subResults
+          unless (null up) $ notifyUP up
+          pure rs
         else do
           logWarn "subcription batch result for replaced SMP client, resubscribing"
           resubscribe $> L.map (second $ \_ -> Left PCENetworkError) rs
       where
-        tSess = transportSession' smp
+        notifyUP up = maybe (logError "sndQ full" >> notifyUP up) pure =<< timeout 30000000 (notifySub c "" $ UP srv up)
```
epoberezkin, line 1369: what is going on here?

dpwiz: `subQ` may become full, blocking everyone up to subscribe command. Happened in the batch test with 200 subs and no reader.
I've added queue queue flushing here, but may be some indication of queue stalling may be helpful to everyone.

PR #1203, opened 2024-06-15
rfc: faster handshake protocol

epoberezkin:


src/Simplex/Messaging/Server/StoreLog.hs
```diff
@@ -53,13 +54,14 @@ data StoreLogRecord
...
-  strEncode QueueRec {recipientId, recipientKey, rcvDhSecret, senderId, senderKey, notifier} =
+  strEncode QueueRec {recipientId, recipientKey, rcvDhSecret, senderId, senderKey, sndSecure, notifier} =
     B.unwords
       [ "rid=" <> strEncode recipientId,
         "rk=" <> strEncode recipientKey,
         "rdh=" <> strEncode rcvDhSecret,
         "sid=" <> strEncode senderId,
-        "sk=" <> strEncode senderKey
+        "sk=" <> strEncode senderKey,
+        "sndSecure=" <> strEncode sndSecure
```
epoberezkin, line 64: possibly, don't save it if it's False to allow server downgrade

PR #1225, opened 2024-07-11
agent: add known servers

epoberezkin:


src/Simplex/Messaging/Agent/Env/SQLite.hs
```diff
@@ -294,3 +330,12 @@ updateRestartCount :: SystemTime -> RestartCount -> RestartCount
 updateRestartCount t (RestartCount minute count) = do
   let min' = systemSeconds t `div` 60
    in RestartCount min' $ if minute == min' then count + 1 else 1
+
+$(pure [])
```
spaced4ndy, line 334: what's this?

src/Simplex/Messaging/Agent/Env/SQLite.hs
```diff
@@ -74,12 +85,37 @@ type AM' a = ReaderT Env IO a
...
+
+data UserServers p = UserServers
+  { enabledSrvs :: NonEmpty (ProtoServerWithAuth p),
+    knownSrvs :: NonEmpty (ProtocolServer p)
+  }
+
+mkUserServers :: NonEmpty (ServerCfg p) -> UserServers p
+mkUserServers srvs = UserServers {enabledSrvs, knownSrvs}
+  where
+    enabledSrvs = L.map (\ServerCfg {server} -> server) $ fromMaybe srvs $ L.nonEmpty $ L.filter (\ServerCfg {enabled} -> enabled) srvs
```
spaced4ndy, line 116: does it make all servers enabled if none are enabled? this is very odd.. even if some UI has checks that prevent it, it's not how I'd expected backend to behave if UI allowed to disable all servers - I think it's expected nothing would work in that case

PR #1232, opened 2024-07-16
agent: allow starting only XFTP sending workers

epoberezkin:


spaced4ndy (dismissed):

src/Simplex/FileTransfer/Agent.hs
```diff
@@ -81,14 +82,16 @@ import UnliftIO
 import UnliftIO.Directory
 import qualified UnliftIO.Exception as E
 
-startXFTPWorkers :: AgentClient -> Maybe FilePath -> AM ()
-startXFTPWorkers c workDir = do
+type AllXFTPWorkers = Bool
```
spaced4ndy, line 85: It's not clear from api which workers start if this is false

PR #1255, opened 2024-08-08
reduce the number of STM transactions

epoberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -2133,7 +2133,7 @@ cleanupManager c@AgentClient {subQ} = do
       step <- asks $ cleanupStepInterval . config
       liftIO $ threadDelay step
     -- we are catching it to avoid CRITICAL errors in tests when this is the only remaining handle to active
-    waitActive a = liftIO (E.tryAny . atomically $ waitUntilActive c) >>= either (\_ -> pure ()) (\_ -> void a)
+    waitActive a = liftIO (E.tryAny $ waitUntilActive c) >>= either (\_ -> pure ()) (\_ -> void a)
```
epoberezkin, line 2136: this is suspicious - why does it need "try"?

spaced4ndy: maybe some older wrong refactoring? not sure. we'd only want to try separate steps and proceed onto next

spaced4ndy (approved):

src/Simplex/Messaging/Agent.hs
epoberezkin: no, waitUntilActive can't throw

epoberezkin: ah - there is some comment...

PR #1273, opened 2024-08-22
Fix the build on OpenBSD

daviduhden:
SimpleXMQ utilizes specific Linux TCP socket options [^1] that are not available in OpenBSD [^2], namely, `TCP_KEEPIDLE`, `TCP_KEEPINTVL`, and `TCP_KEEPCNT`. The absence of these options prevents the successful building of SimpleXMQ on OpenBSD. This change addresses this issue. Additionally, I have attempted to link the binaries with OpenSSL 1.1.1, which is available in the ports tree, by adding the location of these libraries to a `cabal.project.local` file. However, I have been unable to correctly link the binaries with these libraries, encountering persistent errors related to the libraries failing to load. Nevertheless, it is possible to compile SimpleXMQ with the latest version of LibreSSL without any errors.

[^1]: [Linux TCP man page](https://www.man7.org/linux/man-pages/man7/tcp.7.html)
[^2]: [OpenBSD TCP man page](https://man.openbsd.org/tcp)

### DIFF
--- a/src/Simplex/Messaging/Transport/KeepAlive.hs
+++ b/src/Simplex/Messaging/Transport/KeepAlive.hs
@@ -44,20 +44,34 @@ _TCP_KEEPINTVL = 17
 _TCP_KEEPCNT :: CInt
 _TCP_KEEPCNT = 16
 
-#else
--- Mac/Linux
-
-#if defined(darwin_HOST_OS)
+#elif defined(darwin_HOST_OS)
+-- macOS
 foreign import capi "netinet/tcp.h value TCP_KEEPALIVE" _TCP_KEEPIDLE :: CInt
-#else
-foreign import capi "netinet/tcp.h value TCP_KEEPIDLE" _TCP_KEEPIDLE :: CInt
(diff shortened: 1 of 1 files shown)


daviduhden:
The issue with the missing TCP socket options—`TCP_KEEPIDLE`, `TCP_KEEPINTVL`, and `TCP_KEEPCNT`—is specific to the OpenBSD TCP stack. Other UNIX-like operating systems, such as FreeBSD [1], NetBSD [2], DragonFlyBSD [3], and even illumos-based systems [4] (the successor of OpenSolaris), support these options in their TCP stacks.

1. [FreeBSD TCP man page](https://man.freebsd.org/cgi/man.cgi?query=tcp&sektion=4)
2. [NetBSD TCP man page](https://man.netbsd.org/tcp.4)
3. [DragonFlyBSD TCP man page](https://man.dragonflybsd.org/?command=tcp)
4. [illumos TCP man page](https://www.illumos.org/man/4P/TCP)

PR #1290, opened 2024-08-30
newtype for server entity IDs, fix TRcvQueues

epoberezkin:


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -1052,14 +1054,15 @@ sendOrProxySMPCommand ::
...
+  ConnId ->
   ByteString ->
   SMP.SenderId ->
   (SMPClient -> ProxiedRelay -> ExceptT SMPClientError IO (Either ProxyClientError ())) ->
   (SMPClient -> ExceptT SMPClientError IO ()) ->
   AM (Maybe SMPServer)
-sendOrProxySMPCommand c userId destSrv cmdStr senderId sendCmdViaProxy sendCmdDirectly = do
-  sess <- mkTransportSession c userId destSrv senderId
-  ifM shouldUseProxy (sendViaProxy Nothing sess) (sendDirectly sess $> Nothing)
+sendOrProxySMPCommand c userId destSrv connId cmdStr senderId sendCmdViaProxy sendCmdDirectly = do
```
epoberezkin, line 1063: We were creating transport session using sender ID instead of connection ID as usual.

It only affected per-connection transport isolation - additional TCP connections would have been created.

src/Simplex/Messaging/Agent/TRcvQueues.hs
```diff
@@ -113,7 +113,7 @@ isSession rq (uId, srv, connId_) =
 
 instance Queue RcvQueue where
   connId' = connId
-  qKey rq = (userId rq, server rq, connId rq)
+  qKey rq = (userId rq, server rq, rcvId rq)
```
epoberezkin, line 116: This was a proper bug that could have resulted in incorrect subscriptions, particularly when switching connections - both queues would be added under the same key.

PR #1311, opened 2024-09-10
agent: do not reuse notifier credentials when re-creating notification subscription; on deleting ntf sub delete directly on smp server skipping ntf server

spaced4ndy:


src/Simplex/Messaging/Agent/NtfSubSupervisor.hs
```diff
@@ -73,56 +73,34 @@ processNtfSub c (connId, cmd) = do
       case a of
         Nothing -> do
           withTokenServer $ \ntfServer -> do
-            case clientNtfCreds of
-              Just ClientNtfCreds {notifierId} -> do
-                let newSub = newNtfSubscription userId connId smpServer (Just notifierId) ntfServer NASKey
-                withStore c $ \db -> createNtfSubscription db newSub $ NSANtf NSACreate
-                lift . void $ getNtfNTFWorker True c ntfServer
```
spaced4ndy, line 80: unclear to me how this path could occur, possibly dead code

src/Simplex/Messaging/Notifications/Types.hs
```diff
@@ -99,8 +99,8 @@ type NtfActionTs = UTCTime
 data NtfSubNTFAction
   = NSACreate
   | NSACheck
-  | NSADelete
-  | NSARotate
+  | NSADelete -- deprecated
+  | NSARotate -- deprecated
```
epoberezkin, line 103: why deprecated?

src/Simplex/Messaging/Agent/NtfSubSupervisor.hs
```diff
@@ -73,56 +73,34 @@ processNtfSub c (connId, cmd) = do
...
-              withStore' c $ \db -> supervisorUpdateNtfSub db sub (NSANtf NSARotate)
-              lift . void $ getNtfNTFWorker True c subNtfServer
             resetSubscription :: AM ()
             resetSubscription =
               withTokenServer $ \ntfServer -> do
-                let sub' = sub {ntfQueueId = Nothing, ntfServer, ntfSubId = Nothing, ntfSubStatus = NASNew}
+                let sub' = sub {smpServer, ntfQueueId = Nothing, ntfServer, ntfSubId = Nothing, ntfSubStatus = NASNew}
                 withStore' c $ \db -> supervisorUpdateNtfSub db sub' (NSASMP NSASmpKey)
                 lift . void $ getNtfSMPWorker True c smpServer
-    NSCDelete -> do
```
epoberezkin, line 118: don't think you can just remove it, or it'll never reach deleting it on SMP.

At the very least it should be updated to delete on SMP?

spaced4ndy: it's a supervisor command - smp deletion is called instead

src/Simplex/Messaging/Notifications/Types.hs
spaced4ndy: as in no longer used - only keeping them for processing legacy records that may have them saved as action

src/Simplex/Messaging/Agent/NtfSubSupervisor.hs
spaced4ndy: possibly not dead - but reusing for case when creds weren't deleted from queue record - still have to re-create instead of reuse

PR #1312, opened 2024-09-10
smp protocol: send DELD when subscribed queue is deleted

epoberezkin:


src/Simplex/Messaging/Server/QueueStore/STM.hs
```diff
@@ -85,12 +85,14 @@ addQueueNotifier QueueStore {queues, notifiers} rId ntfCreds@NtfCreds {notifierI
       TM.insert nId rId notifiers
       pure q'
 
-deleteQueueNotifier :: QueueStore -> RecipientId -> IO (Either ErrorType ())
+deleteQueueNotifier :: QueueStore -> RecipientId -> IO (Either ErrorType (Maybe NotifierId))
```
spaced4ndy, line 88: in addQueueNotifier (previous function) old notifier creds are also deleted, should it also return notifier id to be written into ntfDeletedQ?

PR #1336, opened 2024-09-25
smp server: persist notifications to avoid losing them when ntf server is offline

epoberezkin:
TODO:
- [x] persist ntfs
- [x] expire ntfs


spaced4ndy (dismissed):

src/Simplex/Messaging/Server/NtfStore.hs
```diff
@@ -1,19 +1,76 @@
...
+      [] -> pure 0
+      _ -> atomically $ do
+        ntfs <- readTVar v
+        case reverse ntfs of
+          -- check the last message first, it is the earliest
+          MsgNtf {ntfTs} : _ | systemSeconds ntfTs < old -> do
+            let !ntfs' = filter (\MsgNtf {ntfTs = ts} -> systemSeconds ts >= old) ntfs
+            writeTVar v ntfs'
+            pure $! length ntfs - length ntfs'
+          _ -> pure 0
```
spaced4ndy, line 65: why not just filter first, then reverse only those written to TVar

smth like

```
let !ntfs' = filter (\MsgNtf {ntfTs = ts} -> systemSeconds ts >= old) ntfs
if null ntfs'
  then
    pure 0
  else do
    writeTVar v (reverse ntfs')
    pure $! length ntfs - length ntfs'
```

that first element is being specially checked here doesn't seem to optimize anything as it requires reversing first anyway

epoberezkin: reverse is much faster operation than filter, and in this case it is executed lazily - that is, it only scans to the last element and takes it - it could just use `last`, as it's known the list is not empty.

PR #1351, opened 2024-10-02
ntf: coalesce wip

spaced4ndy:


epoberezkin:
This cannot be merged until the clients logic to receive multiple messages from a single notification is rolled out.

PR #1369, opened 2024-10-12
smp server: perform major GC periodically

epoberezkin:


src/Simplex/Messaging/Server.hs
```diff
@@ -795,6 +800,15 @@ smpServer started cfg@ServerConfig {transports, transportConfig = tCfg} attachHT
                     logError "Unauthorized control port command"
                     hPutStrLn h "AUTH"
 
+    majorGCThread_ :: ServerConfig -> [M ()]
+    majorGCThread_ ServerConfig {majorGCInterval = Just interval} = [majorGCThread $ interval * 1000000]
+      where
+        majorGCThread int = forever $ do
+          threadDelay int
+          logInfo "Starting major GC"
```
dpwiz, line 808: More like `logDebug`. The message gives even less information than the port bindings, is inactionable yet periodic.

PR #1370, opened 2024-10-12
smp server: journal message store

epoberezkin:

PR #1375, opened 2024-10-21
smp server: journal message store

epoberezkin:
TODO:

- [x] parameterise journal states #1374
- [x] fix write message when restoring messages to collapse quota messages #1380
- [x] validate/expire journal store on start and return correct statistics to use in server stats and to log #1374
- [x] always truncate files on append if necessary, to ensure atomicity (requires adding byteCount to journal states) #1377
- [x] handle read errors (hGetLineAt) #1377
- [x] validate journal state invariants when reading and updating queue state #1377
- [x] enable tests: https://github.com/simplex-chat/simplexmq/pull/1370#discussion_r1808280717

### DIFF
--- a/src/Simplex/Messaging/Server.hs
+++ b/src/Simplex/Messaging/Server.hs
@@ -93,7 +93,7 @@ import Simplex.Messaging.Server.Control
 import Simplex.Messaging.Server.Env.STM as Env
 import Simplex.Messaging.Server.Expiration
 import Simplex.Messaging.Server.MsgStore
-import Simplex.Messaging.Server.MsgStore.Journal (JournalMsgQueue (..), closeMsgQueue)
+import Simplex.Messaging.Server.MsgStore.Journal (JournalMsgQueue (..), JMQueue (..), closeMsgQueue)
 import Simplex.Messaging.Server.MsgStore.STM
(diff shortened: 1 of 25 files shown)


src/Simplex/Messaging/Server.hs
```diff
@@ -1710,59 +1726,115 @@ randomId = fmap EntityId . randomId'
...
+          Just old -> do
+            logInfo "expiring journal store messages..."
+            (storedMsgsCount, expiredMsgsCount, storedQueues) <- withAllMsgQueues ms (\_ -> processExpireQueue old) (0, 0, 0)
+            pure MessageStats {storedMsgsCount, expiredMsgsCount, storedQueues}
+          Nothing -> do
+            logInfo "validating journal store messages..."
+            (storedMsgsCount, storedQueues) <- withAllMsgQueues ms (\_ -> processValidateQueue) (0, 0)
+            pure MessageStats {storedMsgsCount, expiredMsgsCount = 0, storedQueues}
+        where
+          processExpireQueue old q (!stored, !expired, !qCount) =
```
dpwiz, line 1767: Run with MessageStats as accumulator instead of (0,0,0) and enjoy StrictData

epoberezkin: ok

PR #1380, opened 2024-10-22
smp server: merge quota messages and set queue to "over quota" state after restoring, server tests with journal and memory store

epoberezkin:


src/Simplex/Messaging/Server/MsgStore/Journal.hs
```diff
@@ -582,14 +595,6 @@ deleteMsgQueue_ st rId =
   atomically (TM.lookupDelete rId (msgQueues st))
     >>= mapM (\q -> closeMsgQueue q $> state q)
 
-closeMsgQueue :: JournalMsgQueue -> IO ()
```
epoberezkin, line 585: moved to class

PR #1395, opened 2024-10-29
smp server: combine messages and queue storage to optimise performance, prevent race condition when deleting queue and to avoid "orphan" messages.

epoberezkin:


src/Simplex/Messaging/Server/QueueStore/STM.hs
```diff
@@ -1,120 +1,196 @@
...
+    $>>= \q -> q <$$ withLog "addQueue" st (`logCreateQueue` qr)
+  where
+    add = ifM hasId (pure $ Left DUPLICATE_) $ do
+      q <- mkQueue st qr -- STMQueue lock <$> (newTVar $! Just qr) <*> newTVar Nothing
+      TM.insert rId q $ queues' st
+      TM.insert sId rId $ senders' st
+      forM_ notifier $ \NtfCreds {notifierId} -> TM.insert notifierId rId $ notifiers' st
+      pure $ Right q
+    hasId = or <$> sequence [TM.member rId $ queues' st, TM.member sId $ senders' st, hasNotifier]
+    hasNotifier = maybe (pure False) (\NtfCreds {notifierId} -> TM.member notifierId (notifiers' st)) notifier
```
spaced4ndy, line 64: wasn't checked before - was it a bug?

src/Simplex/Messaging/Server/QueueStore/STM.hs
```diff
@@ -1,120 +1,196 @@
...
+updateQueueTime st sq t = atomically (readQueueRec qr >>= mapM update) $>>= log'
+  where
+    qr = queueRec' sq
+    update q@QueueRec {updatedAt}
+      | updatedAt == Just t = pure (q, False)
+      | otherwise =
+          let !q' = q {updatedAt = Just t}
+           in (writeTVar qr $! Just q') $> (q', True)
+    log' (q, changed)
+      | changed = q <$$ withLog "updateQueueTime" st (\sl -> logUpdateQueueTime sl (recipientId q) t)
```
spaced4ndy, line 134: is it new logic - checking if time hasn't changed? does it affect anything?

src/Simplex/Messaging/Server/QueueStore/STM.hs
epoberezkin: We don't create queues with notifiers, it's only important when restoring. We do plan to create queues with user-supplied notifiers in the future, so added this check.

src/Simplex/Messaging/Server/QueueStore/STM.hs
epoberezkin: no, it prevents multiple writes with the same time in case of race condition.

epoberezkin: we don't use this time anywhere yet.

src/Simplex/Messaging/Server/QueueStore/STM.hs
```diff
@@ -1,120 +1,196 @@
...
+
+deleteQueue' :: STMQueueStore s => s -> RecipientId -> StoreQueue s -> IO (Either ErrorType (QueueRec, Maybe (MsgQueue s)))
+deleteQueue' st rId sq =
+  atomically (readQueueRec qr >>= mapM delete)
+    $>>= \q -> withLog "deleteQueue" st (`logDeleteQueue` rId)
+    >>= bimapM pure (\_ -> (q,) <$> atomically (swapTVar (msgQueue_' sq) Nothing))
+  where
+    qr = queueRec' sq
+    delete q = do
+      writeTVar qr Nothing
```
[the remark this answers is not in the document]
epoberezkin: these comments are present in types

PR #1407, opened 2024-11-26
rfc: moving queue records to journal storage and queue blobs, expiring messages in journal storage

epoberezkin:


rfcs/2024-11-25-queue-blobs-2.md
```diff
@@ -0,0 +1,56 @@
...
+- check the last byte of the file and log EOL if it is not EOL. Probably cleanest approach, but with a small performance cost.
+
+If queue folder is a reference to the queue, it may have one of these files:
+- notifier.id
+- sender.id
+- link.id
+
+These files would contain a one line with the recipient ID of the queue. These files would never change, they can only be deleted when queue is deleted or when notifier/link is deleted.
+
+There is logic in code preventing using the same ID in different contexts, and the ID size is large enough to make any collisions unlikely (192 bits), so with correctly working code the queue folder would either have one of reference files, and nothing else, or the queue and message files from the beginning of this section. But even if the same ID is re-used in different context, it should not cause any problems as file names don't overlap.
```
spaced4ndy, line 52: doesn't seem necessary to mix reference folders and recipient folders, as context is always defined in code. they can be stored in different directories, so there's no ambiguity on usage and 4x less space for lookup

recipient_queues
notifier_references
sender_references
link_references

epoberezkin: the reason is that directories are cached, and creating 4x number of hierarchies would make it much less efficient.

epoberezkin: but it's a valid point - I thought about it

PR #1446, opened 2025-02-03
agent: encrypt messages on delivery

spaced4ndy:


src/Simplex/Messaging/Agent.hs
```diff
@@ -1400,13 +1403,20 @@ enqueueMessageB c reqs = do
           agentMsgStr = smpEncode agentMsg
           internalHash = C.sha256Hash agentMsgStr
           currentE2EVersion = maxVersion e2eEncryptVRange
+      -- TODO [save once] Save single MsgBody / enveloped body agentMsgStr (outside of withStoreBatch ... storeSentMsg).
+      -- TODO             Link all messages (message deliveries) to it, save encryption data per message, e.g. pqEnc_ (intent, not result).
+      -- TODO             'msg_body' field is not nullable - use default empty strings?
+      -- TODO             No need to save specific ratchet state/keys for encryption at this point? - Get them at point of delivery?
```
epoberezkin, line 1409: There are probably two options:
1. save ratchet keys at the point of scheduling delivery - this is probably simpler
2. save ratchet keys on the first delivery attempt - the advantage here is that undelivered messages won't be "wasting" IDs and the recipient has less to catch up with on expired messages. The disadvantage here is that there may need to be some other way to signal skipped messages. Although if we use internal IDs for skipped and ratchet numbering is independent it's probably ok.

In any case, there is no option to avoid saving keys / ratchet IDs / headers etc., because they can't change between attempts.

src/Simplex/Messaging/Agent.hs
```diff
@@ -1380,6 +1380,9 @@ enqueueMessage c cData sq msgFlags aMessage =
   ExceptT $ fmap fst . runIdentity <$> enqueueMessageB c (Identity (Right (cData, [sq], Nothing, msgFlags, aMessage)))
 {-# INLINE enqueueMessage #-}
 
+-- TODO [save once] Parameter to save unencrypted message only once in the database - passed from sendMessagesB
+-- TODO             (which would mean client passed same MsgBody in all MsgReq records).
+-- TODO             Or - separate sendBroadcastMessage api?
```
epoberezkin, line 1385: why does it have to be a parameter, and not just how it is always done?

epoberezkin: or, possibly, depending on the number of connections.

tests/AgentTests/FunctionalAPITests.hs
```diff
@@ -728,13 +728,15 @@ testEnablePQEncryption =
     -- switched to smaller envelopes (before reporting PQ encryption enabled)
     sml <- largeMsg g PQSupportOn
     -- fail because of message size
-    Left (A.CMD LARGE _) <- tryError $ A.sendMessage ca bId PQEncOn SMP.noMsgFlags lrg
+    (6, PQEncOff) <- A.sendMessage ca bId PQEncOn SMP.noMsgFlags lrg
+    get ca ##> ("", bId, MERR 6 (A.CMD {cmdErr = LARGE, errContext = "CryptoLargeMsgError"}))
```
spaced4ndy, line 732: by encrypting late we now learn of LARGE message asynchronously and command itself returns success. this would lead to creation of unsent chat items, though they should be marked with error (red x) afterwards. maybe we should add more or less safe limit for unencrypted size, or test encryption immediately upon scheduling delivery, though it is wasteful for most cases

PR #1448, opened 2025-02-05
smp server: PostgreSQL queue store

epoberezkin:


src/Simplex/Messaging/Server.hs
```diff
@@ -1830,50 +1825,54 @@ processServerMessages StartOptions {skipWarnings} = do
...
-            processValidateQueue q =
-              runExceptT (getQueueSize ms q) >>= \case
-                Right storedMsgsCount -> pure newMessageStats {storedMsgsCount, storedQueues = 1}
-                Left e -> do
-                  logError $ "STORE: processValidateQueue, failed opening message queue, " <> tshow e
-                  exitFailure
-
-importMessages :: forall s. STMStoreClass s => Bool -> s -> FilePath -> Maybe Int64 -> Bool -> IO MessageStats
+        AMS _ SMSJournal ms -> processJournalMessages old_ expire ms
+        -- AMS (SType SQSPostgres SMSJournal) ms -> processJournalMessages old_ expire ms
```
spaced4ndy, line 1832: TODO

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -141,13 +145,75 @@ smpServerCLI_ generateSite serveStaticFiles attachStaticFiles cfgPath logPath =
...
+                "Queue records not exported"
+              ps <- newJournalMsgStore $ PQStoreCfg dbOpts MCConsole
+              sl <- openWriteStoreLog storeLogFilePath
+              Sum qCnt <- foldQueueRecs True (postgresQueueStore ps) $ \rId qr -> logCreateQueue sl rId qr $> Sum (1 :: Int)
+              putStrLn $ "Export completed: " <> show qCnt <> " queues"
+              putStrLn $ case readStoreType ini of
+                Right (ASType SQSPostgres SMSJournal) -> "store_queues set to `database`, update it to `memory` in INI file."
+                Right (ASType SQSMemory _) -> "store_queues set to `memory`, start the server"
+                Left e -> e <> ", configure storage correctly"
+        SCDelete -> undefined
```
spaced4ndy, line 199: TODO?

src/Simplex/Messaging/Server/QueueStore/Postgres.hs
```diff
@@ -0,0 +1,375 @@
...
+            DB.execute
+              db
+              [sql|
+                INSERT INTO msg_notifiers (recipient_id, notifier_id, notifier_key, rcv_ntf_dh_secret)
+                VALUES (?, ?, ?, ?)
+                ON CONFLICT (recipient_id) DO UPDATE
+                SET notifier_id = EXCLUDED.notifier_id,
+                    notifier_key = EXCLUDED.notifier_key,
+                    rcv_ntf_dh_secret = EXCLUDED.rcv_ntf_dh_secret
+              |]
```
spaced4ndy, line 199: why would it fail on duplicate notifier_id? schema doesn't have unique constraint on it.

src/Simplex/Messaging/Server/QueueStore/Postgres.hs
```diff
@@ -0,0 +1,375 @@
...
+  deleteStoreQueue st sq =
+    withQueueRec qr delete
+      $>>= \q -> deleteDB
+      >>= mapM (\_ -> (q,) <$> atomically (swapTVar (msgQueue sq) Nothing))
+    where
+      qr = queueRec sq
+      delete q = do
+        writeTVar qr Nothing
+        TM.delete (senderId q) $ senders st
+        -- forM_ (notifier q) $ \NtfCreds {notifierId} -> TM.delete notifierId $ notifiers st
```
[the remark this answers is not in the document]
epoberezkin: because we only insert in this cache when subscription is created, not when queue is read, so with high probability it won't be there. So by letting it linger until restart in a rare case it's there (it was created in the same server session) we avoid the second update - that was the idea.

PR #1468, opened 2025-02-24
agent, smp server: PostgreSQL connection pool

epoberezkin:


src/Simplex/Messaging/Agent/Store/Postgres/Common.hs
```diff
@@ -8,34 +12,96 @@ module Simplex.Messaging.Agent.Store.Postgres.Common
...
+    prepare db = do
+      void $ PSQL.execute_ db "SET client_min_messages TO WARNING"
+      dbNew <- not <$> doesSchemaExist db schema
+      when dbNew $
+        if createSchema
+          then void $ PSQL.execute_ db $ Query $ "CREATE SCHEMA " <> schema
+          else do
+            let err = "schema " <> safeDecodeUtf8 schema <> " does not exist."
+            logError $ "connectPostgresStore: " <> err
+            throwIO $ userError $ T.unpack err
```
spaced4ndy, line 84: why throwIO replaced exitFailure? it can't operate without schema anyway

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -626,6 +632,9 @@ defaultDBConnStr = "postgresql://smp@/smp_server_store"
 defaultDBSchema :: ByteString
 defaultDBSchema = "smp_server"
 
+defaultDBPoolSize :: Natural
+defaultDBPoolSize = 10
```
spaced4ndy, line 636: why not more? they write maximum concurrent connections is 100 by default, I thought we'd configure 20 at least

src/Simplex/Messaging/Agent/Store/Postgres/Common.hs
epoberezkin: because connections are now created while server runs, some sporadic failure shouldn't kill the server.

Maybe a better approach is to create all connections in advance, before starting the servers...

src/Simplex/Messaging/Server/Main.hs
epoberezkin: it's in INI file

PR #1482, opened 2025-03-14
smp server: split postgres support to a separate executable, to not require postgres library in the main binary

epoberezkin:


simplexmq.cabal
```diff
@@ -72,6 +72,11 @@ flag client_postgres
   manual: True
   default: False
 
+flag server_postgres
+  description: Build server without support of PostgreSQL.
```
spaced4ndy, line 76:
```suggestion
  description: Build server with support of PostgreSQL.
```

PR #1483, opened 2025-03-14
smp server: do not cache all queues from database while processing expirations

epoberezkin:
It no longer caches all queues while expiring, but it still creates locks for all queues - so further improvement possible.


src/Simplex/Messaging/Server/MsgStore/Types.hs
```diff
@@ -126,8 +127,11 @@ deleteExpiredMsgs st q old =
 -- returns (expired count, queue size after expiration)
 idleDeleteExpiredMsgs :: MsgStoreClass s => Int64 -> s -> StoreQueue s -> Int64 -> ExceptT ErrorType IO (Maybe Int, Int)
 idleDeleteExpiredMsgs now st q old =
-  isolateQueue q "idleDeleteExpiredMsgs" $
-    withIdleMsgQueue now st q (deleteExpireMsgs_ old q)
+  isolateQueue q "idleDeleteExpiredMsgs" $ do
+    -- Use cached queue if available.
+    -- Also see the comment in cachedOrLoadedQueue
```
spaced4ndy, line 132:
```suggestion
    -- Also see the comment in loadQueue
```

PR #1485, opened 2025-03-17
rfc: smp queue protocol changes

epoberezkin:


rfcs/2025-03-16-smp-queues.md
```diff
@@ -0,0 +1,163 @@
...
+    -- But clients should reject addresses when the primary queue is on a different server.
+    -- Clients could also warn when any of the queues is on unknown operator.
+    userData :: ByteString -- the max size needs to be estimated, but it is likely to be ~ 14kb
+  }
+
+-- | Updated queue IDs and keys, returned in IDS response
+data QueueIdsKeys = QIK
+  { rcvId :: RecipientId,
+    linkId :: Maybe LinkId, -- same as in ShortLink
+    sndId :: SenderId, -- same as in ShortLink or random
```
spaced4ndy, line 81: why do these need to be returned?

rfcs/2025-03-16-smp-queues.md
```diff
@@ -0,0 +1,163 @@
...
+    rcvPublicDhKey :: RcvPublicDhKey,
+    sndSecure :: SenderCanSecure, -- possibly, can be removed? or implied?
+    serverNtfCreds :: Maybe ServerNtfCreds -- currently returned in NID response
+  }
+
+data ServerNtfCreds = ServerNtfCreds NotifierId RcvNtfPublicDhKey -- same NotifierId as in NtfRequest
```
spaced4ndy, line 87: same

rfcs/2025-03-16-smp-queues.md
```diff
@@ -0,0 +1,163 @@
...
+**Prepare queue link data**
+
+- the data blob owner generates X25519 key pair: `(k, pk)`.
+- private key `pk` will be included in the short link shared with the other party (only base64url encoded key bytes, not X509 encoding).
+- `HKDF(pk)` will be used to encrypt the link data with secret_box before storing it on the server.
+- the hash of public key `sha256(k)` will be used as ID by the owner to update the link data and by the accepting party to get link (`LSET` and `LGET` commands).
+
+**Retrieving queue link data**
+
+- the sender uses the public key `k` derived from the private key `pk` included in the link as entity ID to retrieve link data (the server will compute the ID used by the owner as `sha256(k)` and will be able to look it up). This provides the quality that the traffic of the parties has no shared IDs inside TLS. It also means that unlike message queue creation, the ID to retrieve the link data was never sent to the creator, and also is not known to the server in advance (the second part is only an observation, in itself it does not increase security, as server has access to an encrypted blob anyway).
```
spaced4ndy, line 116: the first part also does not increase security, as way to compute id from key is known

rfcs/2025-03-16-smp-queues.md
```diff
@@ -0,0 +1,163 @@
...
+
+```abnf
+shortConnectionRequest = %s"https://" smpServerHost "/" connReqType "#/" linkKey [ "?" param *( "&" param ) ]
+smpServerHost = <hostname> ; RFC1123, RFC5891
+connReqType = %s"i" / %s"invitation" / %s"c" / %s"contact"
+linkKey = <base64url encoded private key data>
+
+param = fingerprintParam / hostsParam
+fingerprintParam = "f=" fingerprint ; server fingerprint
+hostsParam = "h=" <hostname> *( "," <hostname> ) ; additional hostnames, e.g. onion
```
spaced4ndy, line 140: onion address to not be comma separated from smpServerHost? do we allow onion-only addresses currently? (I think yes - then the only address will be in the end)

rfcs/2025-03-16-smp-queues.md
```diff
@@ -0,0 +1,163 @@
...
+https://smp8.simplex.im/c/#abcdefghij0123456789abcdefghij0123456789abc
+```
+
+For the above to work in the browser the servers should serve server pages.
+
+## Questions
+
+SenderId determination:
+- random, as proposed. Pros: no linkability of observed link to queue. Cons: it can be different from the queue, without the accepting party knowing. Potentially, an attack vector?
+- derived from LinkId, and not changeable: Pros: accepting party can verify. Cons: observed link is linkable to the queue.
```
spaced4ndy, line 157: doesn't matter as recipient server is chosen (can be self-hosted) by recipient and can run arbitrary code to run this attack anyway?

rfcs/2025-03-16-smp-queues.md
```diff
@@ -0,0 +1,163 @@
...
+
+Possibly, we could split data blobs to mutable and immutable parts, where SenderId would be immutable, so while it preserves constant SenderId it remains unlinkable.
+
+On another hand, having mutable sender ID may allow delegating the reception of messages to chat relay - when initially messages are received by the user, different contact request URI could point to another SMP client of chat relay, similar to how we plan to do with the group links.
```
spaced4ndy, line 161: didn't get this.. as in allow that client to subscribe with same key / changed key?

rfcs/2025-03-16-smp-queues.md
epoberezkin: so that the difference is contained in encoding layer, and not that it sends completely different responses.

rfcs/2025-03-16-smp-queues.md
epoberezkin: which part?

rfcs/2025-03-16-smp-queues.md
epoberezkin: if it's onion-only it will be in smpServerHost: https://abcd.onion/i#/abcd

rfcs/2025-03-16-smp-queues.md
epoberezkin: The connection request inside the link can be created/controlled by chat relay. That would require some e2e via chat relay, but its parameters could be part of the blob, together with the link.

epoberezkin:
Another question is whether connReq inside blob should be allowed to be shortlink. This can be beneficial to allow multiple connreqs to multiple chat relays, so they take less space. In which case it should control max redirect limit and also prevent circular redirects (possibly via max depth too).

rfcs/2025-03-16-smp-queues.md
```diff
@@ -46,118 +70,186 @@ data NewQueueRequest = NewQueueRequest
...
+- access unencrypted link data, whether it was or was not accessed by the accepting party, provided it has no link (objective 2).
+- observe IP addresses of the users accessing link data, if private routing is used.
+
+**Passive observer who observed short link**:
+
+can:
+- access original unencrypted link data for contact address links.
+
+cannot:
+- undetectably access observed link data, accessing the link would make the link inaccessible to the sender (objective 5).
```
spaced4ndy, line 188:
```suggestion
- undetectably access observed 1-time link data, accessing the link would make the link inaccessible to the sender (objective 5).
```

PR #1495, opened 2025-03-26
smp protocol: short links and improved notifications

epoberezkin:


src/Simplex/Messaging/Agent.hs
```diff
@@ -340,10 +342,27 @@ deleteConnectionsAsync c waitDelivery = withAgentEnv c . deleteConnectionsAsync'
...
+setConnShortLink :: AgentClient -> ConnId -> ConnInfo -> AE (ConnShortLink 'CMContact)
+setConnShortLink c = withAgentEnv c .: setConnShortLink' c
+{-# INLINE setConnShortLink #-}
+
+-- | Get and verify data from short link. For 1-time invitations it preserves the key to allow retries
+getConnShortLink :: AgentClient -> UserId -> ConnShortLink c -> AE (ConnectionRequestUri c, ConnInfo)
+getConnShortLink c = withAgentEnv c .: getConnShortLink' c
+{-# INLINE getConnShortLink #-}
+
+-- | This irreversible deletes short link data, and it won't be retrievable again
```
spaced4ndy, line 360:
```suggestion
-- | This irreversiby deletes short link data, and it won't be retrievable again
```

src/Simplex/Messaging/Agent.hs
```diff
@@ -776,12 +795,45 @@ switchConnectionAsync' c corrId connId =
...
+setConnShortLink' = undefined
+
+getConnShortLink' :: forall c. AgentClient -> UserId -> ConnShortLink c -> AM (ConnectionRequestUri c, ConnInfo)
+getConnShortLink' c userId = \case
+  CSLInvitation srv linkId linkKey -> do
+    g <- asks random
+    invLink <- withStore' c $ \db -> do
+      getInvShortLink db srv linkId >>= \case
+        Just sl@InvShortLink {linkKey = lk} | linkKey == lk -> pure sl
+        _ -> do
```
spaced4ndy, line 815: if it's nothing it should be created, and if key is different it's some internal error and it should be cleaned up?

src/Simplex/Messaging/Agent.hs
```diff
@@ -793,28 +845,86 @@ changeConnectionUser' c oldUserId connId newUserId = do
   where
     updateConn = withStore' c $ \db -> setConnUserId db oldUserId connId newUserId
 
-newRcvConnSrv :: AgentClient -> UserId -> ConnId -> Bool -> SConnectionMode c -> Maybe CRClientData -> CR.InitialKeys -> SubscriptionMode -> SMPServerWithAuth -> AM (ConnectionRequestUri c)
-newRcvConnSrv c userId connId enableNtfs cMode clientData pqInitKeys subMode srvWithAuth@(ProtoServerWithAuth srv _) = do
+newRcvConnSrv :: ConnectionModeI c => AgentClient -> UserId -> ConnId -> Bool -> SConnectionMode c -> Maybe CRClientData -> CR.InitialKeys -> SubscriptionMode -> SMPServerWithAuth -> AM (ConnectionRequestUri c)
```
spaced4ndy, line 848: not sure it's worth to multiply functions for a couple call sites

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -1333,19 +1357,42 @@ getSessionMode :: MonadIO m => AgentClient -> m TransportSessionMode
 getSessionMode = fmap sessionMode . getNetworkConfig
 {-# INLINE getSessionMode #-}
 
-newRcvQueue :: AgentClient -> UserId -> ConnId -> SMPServerWithAuth -> VersionRangeSMPC -> SubscriptionMode -> SenderCanSecure -> AM (NewRcvQueue, SMPQueueUri, SMPTransportSession, SessionId)
-newRcvQueue c userId connId (ProtoServerWithAuth srv auth) vRange subMode senderCanSecure = do
+-- TODO [short links] add ntf credentials to RcvQueue
+newRcvQueue :: AgentClient -> UserId -> ConnId -> SMPServerWithAuth -> VersionRangeSMPC -> SConnectionMode c -> SubscriptionMode -> Maybe NewNtfCreds -> AM (NewRcvQueue, SMPQueueUri, SMPTransportSession, SessionId)
```
spaced4ndy, line 1361: same, newRcvQueue and newRcvQueue_ share 3 call sites excluding calling one another, why not prepare at call site, or make separate function to do preparation from newRcvQueue to be passed to newRcvQueue_ (old newRcvQueue)

src/Simplex/Messaging/Agent.hs
epoberezkin: The key may only be different if the short link is removed from server and then created again. But that would require server re-using the same link_id that should not happen, practically. But it's outside of client control. So it's not exactly internal error. In any case, it would recover from this case because of ON CONFLICT (host, port, link_id) in createInvShortLink.

src/Simplex/Messaging/Agent.hs
epoberezkin: ok

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -1076,9 +1086,24 @@ sendOrProxySMPCommand ::
   (SMPClient -> ProxiedRelay -> ExceptT SMPClientError IO (Either ProxyClientError ())) ->
   (SMPClient -> ExceptT SMPClientError IO ()) ->
   AM (Maybe SMPServer)
-sendOrProxySMPCommand c userId destSrv@ProtocolServer {host = destHosts} connId cmdStr senderId sendCmdViaProxy sendCmdDirectly = do
+sendOrProxySMPCommand c userId destSrv connId cmdStr entId sendCmdViaProxy sendCmdDirectly =
+  fst <$> sendOrProxySMPCommand_ c userId destSrv connId cmdStr entId sendCmdViaProxy sendCmdDirectly
+{-# INLINE sendOrProxySMPCommand #-}
+
+sendOrProxySMPCommand_ ::
```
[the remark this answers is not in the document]
epoberezkin: ok

src/Simplex/Messaging/Agent/Client.hs
epoberezkin: the difference is not so trivial here, so even with 2 call sites better to keep these two extra lines in newRcvQueue here than outside.

PR #1503, opened 2025-04-01
agent: binary protocol encoding for connection request

epoberezkin:


src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -1170,7 +1202,7 @@ data SMPQueueInfo = SMPQueueInfo {clientVersion :: VersionSMPC, queueAddress ::
 
 instance Encoding SMPQueueInfo where
   smpEncode (SMPQueueInfo clientVersion SMPQueueAddress {smpServer, senderId, dhPublicKey, queueMode})
-    | clientVersion >= shortLinksSMPClientVersion = addrEnc <> smpEncode queueMode
+    | clientVersion >= shortLinksSMPClientVersion = addrEnc <> maybe "" smpEncode queueMode
```
epoberezkin, line 1205: this is indeed to avoid potential conflict/overlap with subsequent Maybe in ConnectionRequestUri encoding.

PR #1505, opened 2025-04-03
agent: padded encryption for link data, shorten/restore short links, tests

epoberezkin:


rfcs/2025-04-04-short-links-for-groups.md
```diff
@@ -0,0 +1,47 @@
...
+3. binding owners signatures to the link.
+4. allowing to add/remove owners, both to share ownership and for reliability in case of one owner losing keys/access.
+
+While current short links solve problems 1-3 (via contact type, and via extension of user data in the link), the problem 4 is solved only partially.
+
+We could include the current list of root owners in the user data, and we could send any history of ownership changes from this baseline as a short blockchain on joining the group, we still requrie one master owner to retain access to the queue associated with the group.
+
+## Possible solution approaches
+
+1. "Kick this can down the road" - ignore this problem until there is a namespace, and a group name can be associated with multiple queues.
```
spaced4ndy, line 17: can you clarify this? what namespace?

src/Simplex/Messaging/Crypto/ShortLink.hs
```diff
@@ -18,6 +29,13 @@ import Simplex.Messaging.Agent.Protocol
 import qualified Simplex.Messaging.Crypto as C
 import Simplex.Messaging.Encoding
 import Simplex.Messaging.Protocol (EntityId (..), LinkId, EncDataBytes (..), QueueLinkData)
+import Simplex.Messaging.Util (liftEitherWith)
+
+fixedDataPaddedLength :: Int
+fixedDataPaddedLength = 2008 -- 2048 - 24 (nonce) - 16 (auth tag)
+
+userDataPaddedLength :: Int
+userDataPaddedLength = 13784 -- 13824 - 24 - 16
```
spaced4ndy, line 38: where do 2048 and 13824 come from?

rfcs/2025-04-04-short-links-for-groups.md
epoberezkin: doesn't matter - DNS/decentralized - the one that points a name to multiple queues.

src/Simplex/Messaging/Crypto/ShortLink.hs
epoberezkin: 1-time invitation link with PQ keys is ~1500 bytes, if there will be redundancy here or some other part in fixed data - it reserves ~500 bytes.

contact address is ~130 bytes, but it'll have to be re-worked in the future and will be the same.

epoberezkin: 13824 - just some round number (13.5kb), leaving ~200 bytes for notification credentials.

They are easy to change if some data needs to be added, as the receiving end will accept any size.

It also means that we need to compress profiles.

PR #1525, opened 2025-04-25
ntf server: allow retries when creating subscriptions, prohibit subscriptions with the same queue but another notifier key or token

epoberezkin:


src/Simplex/Messaging/Notifications/Server.hs
```diff
@@ -735,14 +735,13 @@ client NtfServerClient {rcvQ, sndQ} NtfSubscriber {newSubQ, smpAgent = ca} NtfPu
...
-      NtfReqCmd SSubscription (NtfSub NtfSubRec {smpQueue = SMPQueueNtf {smpServer, notifierId}, notifierKey = registeredNKey, subStatus}) (corrId, subId, cmd) -> do
+      NtfReqCmd SSubscription (NtfSub NtfSubRec {ntfSubId, smpQueue = SMPQueueNtf {smpServer, notifierId}, notifierKey = registeredNKey, subStatus}) (corrId, subId, cmd) -> do
         (corrId,subId,) <$> case cmd of
           SNEW (NewNtfSub _ _ notifierKey) -> do
             logDebug "SNEW - existing subscription"
-            -- possible improvement: retry if subscription failed, if pending or AUTH do nothing
             pure $
               if notifierKey == registeredNKey
-                then NRSubId subId
+                then NRSubId ntfSubId
```
epoberezkin, line 744: this was an old bug that returned empty subscription ID on retry, that would probably trigger failure on SCHK, that would then trigger creating a new notifier credentials and new subscription.

PR #1534, opened 2025-05-05
rfc: client certificates for servers using SMP protocol as clients (opertors' chat relays, notification servers, service bots)

epoberezkin:
**SMP server**:
- handshake:
  - [x] request and optionally pass client certificate to SMP server
  - [x] [SMP handshake](https://github.com/simplex-chat/simplexmq/pull/1534/files#diff-c0a195eb9e728c9295a9e4164c2abb23c11a983973991e38c36b766e153ed01aR693) with additional server response 8558f2fa5cd7c0ddff013c9426eaa85bfc57a32b
  - [x] get/create service record in handshake (currently, there is no database access in handshake, probably should be done via hook).
- verification
  - [x] validate that command is allowed for service/role combination https://github.com/simplex-chat/simplexmq/pull/1534/commits/9734edc8c511291208bbd8cb3ad322d322ae50cb
  - [x] verify service signature https://github.com/simplex-chat/simplexmq/pull/1534/commits/9734edc8c511291208bbd8cb3ad322d322ae50cb
- process [server commands](https://github.com/simplex-chat/simplexmq/pull/1534/files#diff-0e388a1f4f15b6437f096ceb808503f15a7a9dc0a754b6a91772fc85ed2d35ceR418):
  - associate messaging service ID with queue
    - [ ] NEW command
    - [ ] SUB command (change response to SOK in all cases)
    - [x] NSUB command (change response to SOK)
  - [ ] process SSUB command
  - [x] process NSSUB command
  - [x] process subscribed client/service changes in serverThread
- deliver events to connected service, based on associated service ID:
  - [ ] messages
  - [x] notifications
  - [x] DELD
  - [x] END to individually subscribed connections when service subscribes, and vice versa
- statistics
  - [x] encode/decode
  - [x] include in prometheus
- storage
  - [x] stm
  - [x] postgres

**Notification server**:
- initial subscription:
  - [x] handover to service (association).
  - [x] re-handover if service ID changes in handshake or service becomes unavailable.
  - [x] service subscription for queues that have the same service ID as in handshake.
- [x] service subscriptions workers in "small agent"
- [x] create association with service ID in CASubscribed
- [x] additional "small agent" events when service subscriptions succeed/end
- [x] prometheus: track counts of subscribed queues in service subscriptions

**Chat Agent**:
- [ ] network setting to create certificates for the service, one per user+server combination.
- [ ] send certificate in TLS/SMP handshakes, store service ID.
- create association:
  - [ ] creating queue (newRcvQueue) when creating connection, joining and switching connections
  - [ ] response to subscriptions.
- [ ] return service association in [async responses](https://github.com/simplex-chat/simplexmq/pull/1534/files#diff-7e590efc6d146da6ccf060599aea1aee68edce0f8ff35452a964bce93d25f795R375).
- [ ] service subscription [agent API](https://github.com/simplex-chat/simplexmq/pull/1534/files#diff-a3e285d7ba89c10719fd33b2e5aa74bd5c6d3340750465d0179c108dd3fa18bdR370) - should accept user and service ID, respond with error if handshake reports different service ID, so that chat can re-handover.
- [ ] account for service subscriptions in agent stats

**Chat Core**:
- [ ] probably, should store service association on connections too.
- [ ] maintain the set of subscribed services, report them to UI, so that spinners disappear.
- [ ] initial subscription - read services for all connections, read unassociated connections only to handover.
- [ ] re-handover on agent error (if service ID changes).


rfcs/2025-05-05-client-certificates.md
```diff
@@ -0,0 +1,129 @@
...
+
+The absense of user and client identification benefits privacy, but it requires separately authorizing subscription for each messaging queue, that simply doesn't scale even for the current traffic and network size.
+
+While chat relays (aka super-peers) would reduce the number of subscriptions required for the usual clients, by replacing connections with each group member to 1-3 connections with chat relays per group/community, it would shift the burden to the chat relays, that are also clients.
+
+While self-hosted chat relays may want to retain privacy, this is not needed (and counter-productive) for the chat relays provided by network operators.
+
+Even today, directory service subscribing to all queues may take 15-20 minutes, which is experienced as downtime by the end users.
+
+Notification servers are also, effectively, clients that subscribe to the messaging servers, and it also takes 15-20 minutes.
```
spaced4ndy, line 13: it is planned to change notifications protocol from subscription to push though, so it would stop to be a part of a problem

rfcs/2025-05-05-client-certificates.md
```diff
@@ -0,0 +1,129 @@
...
+
+`ServerHandshake` already contains `authPubKey` with the server certificate chain and the signed key for connection encryption and creating a shared secret for denable authorization (with client entity key) and session encryption layer.
+
+`ClientHandshake` contains only ephemeral `authPubKey` to compute a shared secret for session encryption layer, so we need an additional field for client certificate:
+
+```haskell
+clientPubKey :: Maybe X.CertificateChain
+```
+
+For operators' clients we may optionally include operators' certificate in the chain, and that would allow servers to identify operators if either wants to. This would improve end-user security, as not only the server would validate that its certificate matches the address, but it would also validate that it is operated by SimpleX Chat or by Flux, preventing any server impersonation (e.g., via DNS manipulations) - the client could then report that the files are hosted on SimpleX Chat servers, but then can stop and show additional warning in case certificate does not match the domain - same as the browsers do with CA stores in the client.
```
spaced4ndy, line 74: > ... preventing any server impersonation ...

meaning preventing any operator client impersonation?

rfcs/2025-05-05-client-certificates.md
```diff
@@ -0,0 +1,129 @@
...
+CSQS :: Word32 -> BrokerMsg -- response from the server, includes the number of subscribed queues
+```
+
+## Ephemeral client-session association
+
+This was considered to reduce costs for the usual clients to re-subscribe. Currently it's a big problem, because of groups, and with transition to chat relays it won't be.
+
+For some very busy end-user clients it may help.
+
+Given that server has access to an ephemeral association between recipient client session and queues anyway (even with clients connecting via Tor, unless per-connection transport isolation is used), introducing `sessionPubKey` to allow resubscription to the previously subscribed queues may reduce the traffic. This won't change threat model as the server would only keep this association in memory, and not persist it. Clients on another hand may safely persist this association for fast resubscription on client restarts.
```
spaced4ndy, line 123: server keeping it in memory seems more like a promise and not hard thread model

rfcs/2025-05-05-client-certificates.md
epoberezkin: Yes, but it's a more complex change, and we have 10s of millions of existing subscriptions to migrate anyway.

It is also necessary for directory and for chat relays.

rfcs/2025-05-05-client-certificates.md
```diff
@@ -0,0 +1,129 @@
...
+All queues created with client key will have to be double-authorized with both the queue key and the client key - both the client and the server would have to maintain this knowledge, whether the queue is associated with the client or not.
+
+Asymmetric retries have to be considered - the first request creating this association may succeed on the server and timeout on the client.
+
+## Subscription
+
+There are two options:
+
+- automatically subscribe client on connection in case certificate is introduced. This avoids additional command for the client, but it may work for considered ephemeral session associations with the usual clients (see below) for iOS NSE clients.
+- send a separate command authorized with the client key passed in handshake.
```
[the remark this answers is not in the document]
epoberezkin: possibly, yes. But I don't see why it's needed with the proposed design.

rfcs/2025-05-05-client-certificates.md
epoberezkin: "threat". Yes. But the point is that ephemeral session certificates don't change it in any way, as the server can track queues that clients subscribe to.

As for "hardness" of promise, TEEs can make it harder.

rfcs/2025-05-05-client-certificates.md
epoberezkin: and btw I don't think it will be pushed. Subscription creation needs to change to push, but actual subscription is better to continue using this flow when ntf server connects and presents its certificate.

something to think about.

src/Simplex/Messaging/Server/QueueStore/Postgres/Migrations.hs
```diff
@@ -119,3 +120,42 @@ UPDATE msg_queues SET recipient_keys = substring(recipient_keys from 3);
...
+CREATE TABLE services(
+  service_id BYTEA NOT NULL,
+  service_role TEXT NOT NULL,
+  service_cert BYTEA NOT NULL,
+  service_cert_hash BYTEA NOT NULL UNIQUE,
+  created_at BIGINT NOT NULL,
+  PRIMARY KEY (service_id)
+);
+
+CREATE INDEX idx_services_service_role ON services(service_role);
```
epoberezkin, line 137: where it's used?

src/Simplex/Messaging/Transport.hs
```diff
@@ -450,22 +475,37 @@ data THandleParams v p = THandleParams
...
+    -- | include service signature (or '0' if it is absent), based on protocol version
+    serviceAuth :: Bool
   }
 
 data THandleAuth (p :: TransportPeer) where
   THAuthClient ::
-    { serverPeerPubKey :: C.PublicKeyX25519, -- used by the client to combine with client's private per-queue key
-      serverCertKey :: CertChainPubKey, -- the key here is serverPeerPubKey signed with server certificate
+    { peerServerPubKey :: C.PublicKeyX25519, -- used by the client to combine with client's private per-queue key
+      peerServerCertKey :: CertChainPubKey, -- the key here is serverPeerPubKey signed with server certificate
```
spaced4ndy, line 486:
```suggestion
      peerServerCertKey :: CertChainPubKey, -- the key here is peerServerPubKey signed with server certificate
```

PR #1573, opened 2025-06-17
smp server: refactor subscriptions and delivery

epoberezkin:


src/Simplex/Messaging/Server.hs
```diff
@@ -1854,7 +1870,7 @@ client
                       ProhibitSub -> pure Nothing
                       ServerSub st -> readTVar st >>= \case
                         NoSub ->
-                          tryTakeTMVar delivered >>= \case
+                          tryReadTMVar delivered >>= \case
```
epoberezkin, line 1873: This looks like an incorrect logic that had no effect as it _should not be_ possible to have concurrent deliveries to an empty queue (see `wasEmpty` condition above). But if it were somehow possible, it would have resulted in stuck delivery.

src/Simplex/Messaging/Server.hs
```diff
@@ -1878,9 +1894,9 @@ client
                       -- lookup can be outside of STM transaction,
                       -- as long as the check that it is the same client is inside.
                       getSubscribedClient rId (queueSubscribers subscribers) >>= mapM_ deliverIfSame
-                    deliverIfSame rcv = time "deliver" . atomically $
+                    deliverIfSame rcv = atomically $
                       whenM (sameClient rc rcv) $
-                        tryTakeTMVar delivered >>= \case
+                        tryReadTMVar delivered >>= \case
```
epoberezkin, line 1899: see comment above

src/Simplex/Messaging/Server.hs
```diff
@@ -1957,21 +1974,6 @@ client
                     VRVerified q -> Right (q, t'')
                     VRFailed e -> Left (corrId', entId', ERR e)
 
-        deliverMessage :: T.Text -> QueueRec -> RecipientId -> Sub -> Maybe Message -> IO (Transmission BrokerMsg)
```
epoberezkin, line 1960: this function unnecessarily duplicates the check for ProhibitSub at call sites and prevents different usage of messages in SUB and in ACK commands.

src/Simplex/Messaging/Server.hs
```diff
@@ -1990,11 +1992,11 @@ client
...
               -- Possibly, the same should be done if the queue is suspended, but currently we do not use it
-              atomically $ do
-                writeTQueue (subQ subscribers) (CSDeleted entId rcvServiceId, clientId)
-                -- queue is usually deleted by the same client that is currently subscribed,
-                -- we delete subscription here, so the client with no subscriptions can be disconnected.
-                TM.delete entId subscriptions
+              -- queue is usually deleted by the same client that is currently subscribed,
+              -- we delete subscription here, so the client with no subscriptions can be disconnected.
+              sub <- atomically $ TM.lookupDelete entId $ subscriptions clnt
+              liftIO $ mapM_ cancelSub sub
```
epoberezkin, line 1998: the lack of cancel was potentially a memory leak

PR #1588, opened 2025-07-16
rfc: using the same profile from multiple devices

epoberezkin:


rfcs/2025-07-15-multi-device.md
```diff
@@ -0,0 +1,104 @@
...
+- manipulating information by sending messages to one device (to have proof it was sent) and withholding from others, or sending different messages if the protocol allows it.
+
+In addition to that, the specific implementation of this approach in Signal compromises break-in recovery property (aka post-compromise security) of Double-Ratchet algorithm, making its design ineffective - the only reason to have the second ratchet in DR algorithm is to provide break-in recovery, without it a much simpler design with a single ratchet is sufficient. See [this paper](https://eprint.iacr.org/2021/626.pdf) for details.
+
+While this limitation can be addressed with notifications when a new device is added and per-device keys, we still find the remaining attack vectors on user security and privacy to be unacceptable, and opening unsuspecting users to various criminal actions - and it is wrong to say that would only affect security conscious users, and most people would not be affected by these risks. Allowing potential criminals in groups to know which device you are currently using is a real risk for all users.
```
spaced4ndy, line 13: I think this prematurely throws away this approach in favor of more fancy algorithm, because status quo is even worse - no convenience and the approach we unofficially recommend (add your devices to groups) is as bad, if not worse, since it's even easier to differentiate user devices, even without client modifications.

Maybe at least cost of this approach should be considered. If it is much cheaper than what rfc proposes it would actually improve privacy for users who do add devices (effort of modifying client is required) and greatly boost convenience, and possibly it could happen / be prioritized sooner.

Another point is that in large groups chat relays could mask user devices, so that'd be one less downside. And yes while it'd still be visible to chat relays, it's server trust issue, and in rfc protocol servers also can differentiate devices, though only on a lower protocol level.

rfcs/2025-07-15-multi-device.md
```diff
@@ -0,0 +1,104 @@
...
+
+1. Server sends the message to the device (can be in response to SUB or ACK commands, or with active subscription). Pushed message would include the hash of the currently stored ratchet state: `hash(enc(rs))`.
+
+2. If device has the ratchet state with the same hash (`hash(enc(rs)) == hash(enc(rsi))`), it decrypts the message: `(pt, rsi') = ratchetDec(ct, rsi)`.
+
+3. If device has ratchet state with a different hash, it requests ratchet from the server with additional protocol command `RGET` with response `RCHT (enc(rs))` and updates the local state.
+
+4. Device decrypts the message `(pt, rsi') = ratchetDec(ct, rsi)` and processes it as usual.
+
+5. Device sends acknowledgement to the server as usual, but now it includes the new ratchet state and the hash of the previous state: `ACK msgId (hash(enc(rsi)), enc(rsi'))`
```
spaced4ndy, line 85: same question as above

rfcs/2025-07-15-multi-device.md
epoberezkin: > even easier to differentiate user devices, even without client modifications.

only in chats where user wants it.

rfcs/2025-07-15-multi-device.md
epoberezkin: Repeat ACKs are accepted.

PR #1611, opened 2025-08-27
ntf server: move token functions and types to shared module to be used with WebPush

p1gp1g:
This PR moves a few functions from APNS module to the Push module, as they will be used for WebPush

### DIFF
--- a/simplexmq.cabal
+++ b/simplexmq.cabal
@@ -259,6 +259,7 @@ library
               Simplex.Messaging.Notifications.Server.Main
               Simplex.Messaging.Notifications.Server.Prometheus
               Simplex.Messaging.Notifications.Server.Push.APNS
+              Simplex.Messaging.Notifications.Server.Push
               Simplex.Messaging.Notifications.Server.Push.APNS.Internal
               Simplex.Messaging.Notifications.Server.Stats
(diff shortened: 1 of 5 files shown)


simplexmq.cabal
```diff
@@ -259,6 +259,7 @@ library
               Simplex.Messaging.Notifications.Server.Main
               Simplex.Messaging.Notifications.Server.Prometheus
               Simplex.Messaging.Notifications.Server.Push.APNS
+              Simplex.Messaging.Notifications.Server.Push
```
epoberezkin, line 262:
```suggestion
              Simplex.Messaging.Notifications.Server.Push
              Simplex.Messaging.Notifications.Server.Push.APNS
```

src/Simplex/Messaging/Notifications/Server.hs
```diff
@@ -675,6 +675,7 @@ ntfPush s@NtfPushServer {pushQ} = forever $ do
             void $ updateTknStatus st tkn $ NTInvalid $ Just r
             err e
           PPPermanentError -> err e
+          PPInvalidPusher -> err e
```
epoberezkin, line 678:
```suggestion
```

Let's add it where it's used please (I will add if I see in review)

PR #1612, opened 2025-08-27
Rename ntf DeviceToken APNSDeviceToken

p1gp1g:
This PR is based on #1611

It rename the previous DeviceToken to APNSDeviceToken, in order to introduce WPDeviceToken later

PR #1613, opened 2025-08-27
Webpush requests

p1gp1g:
This PR is based on #1612 

It adds a WebPush push provider and do the web push requests

It doesn't contain the VAPID authorization yet (as the proposed implementation was using extras, that may be dropped)


src/Simplex/Messaging/Notifications/Server/Env.hs
```diff
@@ -159,13 +164,27 @@ newNtfPushServer qSize apnsConfig = do
...
     Just host -> apnsPushProviderClient <$> createAPNSPushClient host apnsConfig
   atomically $ TM.insert pp c pushClients
   pure c
 
+newWPPushClient :: NtfPushServer -> IO PushProviderClient
+newWPPushClient NtfPushServer {pushClients} = do
+  logDebug "New WP Client requested"
+  manager <- newManager tlsManagerSettings
+  let c = wpPushProviderClient manager
+  atomically $ TM.insert PPWebPush c pushClients
```
epoberezkin, line 185: I don't think it'll work, even if the user can only use one push provider, all users can use different push providers, so the server should store in the map all of them, and either the key should be different or push provider should include endpoint (which is probably better, as endpoint is also part of token definition)

epoberezkin: so what would happen is that it would be pushing to a random push server

src/Simplex/Messaging/Crypto.hs
```diff
@@ -1021,11 +1024,22 @@ encryptAESNoPad :: Key -> GCMIV -> ByteString -> ExceptT CryptoError IO (AuthTag
...
+{-# INLINE encryptAES128NoPad #-}
+
 encryptAEADNoPad :: Key -> GCMIV -> ByteString -> ByteString -> ExceptT CryptoError IO (AuthTag, ByteString)
 encryptAEADNoPad aesKey ivBytes ad msg = do
   aead <- initAEADGCM aesKey ivBytes
   pure . first AuthTag $ AES.aeadSimpleEncrypt aead ad msg authTagSize
 
+encryptAEAD128NoPad :: Key -> GCMIV -> ByteString -> ByteString -> ExceptT CryptoError IO (AuthTag, ByteString)
+encryptAEAD128NoPad aesKey ivBytes ad msg = do
+  aead <- initAEAD128GCM aesKey ivBytes
```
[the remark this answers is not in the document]
p1gp1g: Yes: https://www.rfc-editor.org/rfc/rfc8291#section-4

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+      _ <- liftPPWPError $ httpNoBody req mg
+      pure ()
+  where
+    endpoint :: NtfTknRec -> ExceptT PushProviderError IO WPEndpoint
+    endpoint NtfTknRec {token} = do
+      case token of
+        WPDeviceToken e -> pure e
+        _ -> fail "Wrong device token"
+    -- TODO: move to PPIndalidPusher ? WPEndpoint should be invalidated and removed if the key is invalid, but the validation key is never sent
+    body :: WPEndpoint -> ExceptT PushProviderError IO B.ByteString
```
epoberezkin, line 74: we can't use OverloadedRecordDot extension, we need backwards compatibility with 8.10.7. So it should either be unique selectors names (so one can write `auth e`) or NamedFieldsPun extension so it can be WPEndpoint {auth}

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+import qualified Crypto.Error as CE
+import qualified Crypto.MAC.HMAC as HMAC
+import qualified Crypto.PubKey.ECC.DH as ECDH
+import qualified Crypto.PubKey.ECC.Types as ECC
+import GHC.Base (when)
+
+wpPushProviderClient :: Manager -> PushProviderClient
+wpPushProviderClient mg tkn pn = do
+      e <- endpoint tkn
+      r <- liftPPWPError $ parseUrlThrow $ B.unpack e.endpoint
```
epoberezkin, line 50: this parsing needs to happen when token is received, and converted to type that doesn't require parsing at a point of sending notification

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+    body :: WPEndpoint -> ExceptT PushProviderError IO B.ByteString
+    body e = withExceptT PPCryptoError $ wpEncrypt e.auth e.p256dh (BL.toStrict $ encodePN pn)
+
+-- | encrypt :: auth -> key -> clear -> cipher
+-- | https://www.rfc-editor.org/rfc/rfc8291#section-3.4
+wpEncrypt :: B.ByteString -> B.ByteString -> B.ByteString -> ExceptT C.CryptoError IO B.ByteString
+wpEncrypt auth uaPubKS clearT = do
+  salt :: B.ByteString <- liftIO $ getRandomBytes 16
+  asPrivK <- liftIO $ ECDH.generatePrivate $ ECC.getCurveByName ECC.SEC_p256r1
+  uaPubK <- point uaPubKS
```
epoberezkin, line 83: this is effectively parsing, should be done in type parser, not here.

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+import qualified Crypto.Cipher.Types as CT
+import qualified Crypto.Error as CE
+import qualified Crypto.MAC.HMAC as HMAC
+import qualified Crypto.PubKey.ECC.DH as ECDH
+import qualified Crypto.PubKey.ECC.Types as ECC
+import GHC.Base (when)
+
+wpPushProviderClient :: Manager -> PushProviderClient
+wpPushProviderClient mg tkn pn = do
+      e <- endpoint tkn
```
epoberezkin, line 49: that also should not be failing

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+            , ("Content-Encoding", "aes128gcm")
+        -- TODO: topic for pings and interval
+            ]
+          req = r {
+        method = "POST"
+        , requestHeaders
+        , requestBody = RequestBodyBS encBody
+        , redirectCount = 0
+      }
+      _ <- liftPPWPError $ httpNoBody req mg
```
epoberezkin, line 65: this doesn't seem to use TLS

simplexmq.cabal
```diff
@@ -260,6 +260,8 @@ library
               Simplex.Messaging.Notifications.Server.Prometheus
               Simplex.Messaging.Notifications.Server.Push
               Simplex.Messaging.Notifications.Server.Push.APNS
+              Simplex.Messaging.Notifications.Server.Push.WebPush
+              Simplex.Messaging.Notifications.Server.Push
```
epoberezkin, line 264:
```suggestion
```

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -1316,7 +1316,7 @@ runNTFServerTest c@AgentClient {presetDomains} nm userId (ProtoServerWithAuth sr
         (nKey, npKey) <- atomically $ C.generateAuthKeyPair a g
         (dhKey, _) <- atomically $ C.generateKeyPair g
         r <- runExceptT $ do
-          let deviceToken = DeviceToken PPApnsNull "test_ntf_token"
+          let deviceToken = APNSDeviceToken PPApnsNull "test_ntf_token"
```
epoberezkin, line 1319: These renames are unnecessary, as we still need a unified API for all tokens, both APNS and WebPush. See the type for DeviceToken below.

src/Simplex/Messaging/Agent/Store/AgentStore.hs
```diff
@@ -1382,7 +1382,8 @@ deleteCommand db cmdId =
   DB.execute db "DELETE FROM commands WHERE command_id = ?" (Only cmdId)
 
 createNtfToken :: DB.Connection -> NtfToken -> IO ()
-createNtfToken db NtfToken {deviceToken = DeviceToken provider token, ntfServer = srv@ProtocolServer {host, port}, ntfTokenId, ntfPubKey, ntfPrivKey, ntfDhKeys = (ntfDhPubKey, ntfDhPrivKey), ntfDhSecret, ntfTknStatus, ntfTknAction, ntfMode} = do
+createNtfToken db NtfToken {deviceToken, ntfServer = srv@ProtocolServer {host, port}, ntfTokenId, ntfPubKey, ntfPrivKey, ntfDhKeys = (ntfDhPubKey, ntfDhPrivKey), ntfDhSecret, ntfTknStatus, ntfTknAction, ntfMode} = do
+  let (provider, token) = deviceTokenFields deviceToken
```
epoberezkin, line 1386: we most likely need to add fields for webpush tokens to the table.

src/Simplex/Messaging/Crypto.hs
```diff
@@ -895,6 +896,8 @@ data CryptoError
     CERatchetEarlierMessage Word32
   | -- | duplicate message number
     CERatchetDuplicateMessage
+  | -- | unable to decode ecc key
+    CryptoInvalidECCKey CE.CryptoError
```
epoberezkin, line 900:
```suggestion
```

This error is not necessary as it would be a parsing error.

src/Simplex/Messaging/Notifications/Protocol.hs
```diff
@@ -399,44 +403,116 @@ instance StrEncoding PushProvider where
...
-    pure $ DeviceToken pp t
+    case pp of
+      PPWebPush -> do
+        WPDeviceToken <$> (o .: "token")
+      _ -> do
+        t <- encodeUtf8 <$> (o .: "token")
+        pure $ APNSDeviceToken pp t
+
+-- | Returns fields for the device token (pushProvider, token)
+deviceTokenFields :: DeviceToken -> (PushProvider, ByteString)
```
epoberezkin, line 506: this function seems unnecessary, as it's better to store components as separate fields and split/combine when saving to the database.

src/Simplex/Messaging/Notifications/Server/Env.hs
```diff
@@ -160,13 +164,27 @@ newNtfPushServer qSize apnsConfig = do
...
   c <- case apnsProviderHost pp of
     Nothing -> pure $ \_ _ -> pure ()
     Just host -> apnsPushProviderClient <$> createAPNSPushClient host apnsConfig
   atomically $ TM.insert pp c pushClients
   pure c
 
+newWPPushClient :: NtfPushServer -> IO PushProviderClient
+newWPPushClient NtfPushServer {pushClients} = do
+  logDebug "New WP Client requested"
+  manager <- newManager tlsManagerSettings
```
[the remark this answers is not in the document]
epoberezkin: TODO: validate if it uses default system store, most likely it does not.

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
epoberezkin: see comment above

epoberezkin:
Added all (I think:) comments. I probably need to make a change on top with different types.

src/Simplex/Messaging/Notifications/Server/Env.hs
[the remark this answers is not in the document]
p1gp1g: tlsManagerSettings loads certificates from /etc/ssl /system/etc/security/cacerts and /usr/local/share/certs ; if no CA are available, all requests fail

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+-- | Elliptic-Curve-Point-to-Octet-String Conversion without compression
+-- | as required by RFC8291
+-- | https://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3
+uncompressEncode :: ECC.Point -> BL.ByteString
+uncompressEncode (ECC.Point x y) = "\x04" <>
+                                     encodeBigInt x <>
+                                     encodeBigInt y
+uncompressEncode ECC.PointO = "\0"
+
+uncompressDecode :: BL.ByteString -> ExceptT CE.CryptoError IO ECC.Point
```
epoberezkin, line 121: there is no reason for it to be IO, it needs to evaluate to Either

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+encodeBigInt i = do
+  let s1 = Bits.shiftR i 64
+      s2 = Bits.shiftR s1 64
+      s3 = Bits.shiftR s2 64
+  Bin.encode ( w64 s3, w64 s2, w64 s1, w64 i )
+  where
+    w64 :: Integer -> Bin.Word64
+    w64 = fromIntegral
+
+decodeBigInt :: BL.ByteString -> ExceptT CE.CryptoError IO Integer
```
epoberezkin, line 143: there is no reason for it to be IO, it needs to evaluate to Either

src/Simplex/Messaging/Notifications/Server/Push/WebPush.hs
```diff
@@ -0,0 +1,178 @@
...
+decodeBigInt s = do
+  when (BL.length s /= 32) $ throwError CE.CryptoError_PointSizeInvalid
+  let (w3, w2, w1, w0) = Bin.decode s :: (Bin.Word64, Bin.Word64, Bin.Word64, Bin.Word64 )
+  pure $ shift 3 w3 + shift 2 w2 + shift 1 w1 + shift 0 w0
+  where
+    shift i w = Bits.shiftL (fromIntegral w) (64*i)
+
+encodePN :: PushNotification -> BL.ByteString
```
epoberezkin, line 151: this looks like the usual JSON encoding, we should use generated ToJSON instance adding selectors to the type

src/Simplex/Messaging/Notifications/Protocol.hs
```diff
@@ -399,44 +403,116 @@ instance StrEncoding PushProvider where
...
+          PPWebPush -> WPDeviceToken <$> strP
+          _ -> APNSDeviceToken pp <$> hexStringP
       hexStringP =
         A.takeWhile (`B.elem` "0123456789abcdef") >>= \s ->
           if even (B.length s) then pure s else fail "odd number of hex characters"
 
 instance ToJSON DeviceToken where
-  toEncoding (DeviceToken pp t) = J.pairs $ "pushProvider" .= decodeLatin1 (strEncode pp) <> "token" .= decodeLatin1 t
-  toJSON (DeviceToken pp t) = J.object ["pushProvider" .= decodeLatin1 (strEncode pp), "token" .= decodeLatin1 t]
+  toEncoding token = case token of
```
epoberezkin, line 488: why do we need JSON encodings for the token? If we do, then we need to use derived ToJSON/FromJson instances.

Possibly, it is only needed for webpush tokens?

PR #1684, opened 2025-12-25
agent: create user with option to enable client service

epoberezkin:


src/Simplex/Messaging/Transport/HTTP2/Client.hs
```diff
@@ -176,9 +184,9 @@ sendRequest HTTP2Client {client_ = HClient {config, reqQ}} req reqTimeout_ = do
 sendRequestDirect :: HTTP2Client -> Request -> Maybe Int -> IO (Either HTTP2ClientError HTTP2Response)
 sendRequestDirect HTTP2Client {client_ = HClient {config, disconnected}, sendReq} req reqTimeout_ = do
   let reqTimeout = http2RequestTimeout config reqTimeout_
-  reqTimeout `timeout` E.try (sendReq req process) >>= \case
+  reqTimeout `timeout` ((Right <$> sendReq req process) `E.catches` httpClientHandlers) >>= \case
     Just (Right r) -> pure $ Right r
-    Just (Left (e :: E.SomeException)) -> disconnected $> Left (HCIOError $ E.displayException e)
+    Just (Left e) -> disconnected $> Left e
```
epoberezkin: Catching SomeException that previously replaced IOException results in undesirable catching of async exceptions.

We need to review all catch, try etc. from Control.Exceptions and use safe alternatives that don't catch async exceptions. catchAllErrors etc. are already safe as they use UnliftIO.

PR #1784, opened 2026-05-22
smp-server: support namespaces

shumvgolove:


src/Simplex/Messaging/Names/Record.hs
```diff
@@ -0,0 +1,89 @@
...
+-- MULTI_URL_FIELD_CAP = 5); records with more entries are non-conforming.
+maxLinks :: Int
+maxLinks = 5
+
+maxLinkBytes :: Int
+maxLinkBytes = 1024
+
+-- FromJSON is hand-rolled to enforce per-field UTF-8 byte-length caps that
+-- TH derivation cannot express.
+instance J.FromJSON NameRecord where
```
epoberezkin: There is no reason to parse it manually, make a haskell type that matches that shape.

src/Simplex/Messaging/Names/Record.hs
```diff
@@ -0,0 +1,89 @@
...
+    nrXmr <- o J..:? "xmr" >>= traverse (capUtf8 "xmr" 255)
+    nrDot <- o J..:? "dot" >>= traverse (capUtf8 "dot" 255)
+    nrOwner <- o J..: "owner"
+    nrResolver <- o J..: "resolver"
+    pure NameRecord {nrName, nrNickname, nrWebsite, nrLocation, nrSimplexContact, nrSimplexChannel, nrEth, nrBtc, nrXmr, nrDot, nrOwner, nrResolver}
+    where
+      capUtf8 fld lim t
+        | B.length (encodeUtf8 t) <= lim = pure t
+        | otherwise = fail $ fld <> " exceeds " <> show lim <> " bytes UTF-8"
+      capLinks fld links
```
epoberezkin: No reason to have this limit in type/parser

epoberezkin: The only real limit is that it fits in transport packet.

src/Simplex/Messaging/Names/Owner.hs
```diff
@@ -0,0 +1,46 @@
...
+
+-- | 20-byte Ethereum address (NameRecord owner). Bare constructor not exported;
+-- use `mkNameOwner` to enforce the 20-byte invariant.
+newtype NameOwner = NameOwner ByteString
+  deriving (Eq)
+
+-- Render the 20 raw bytes as "0x"-prefixed lowercase hex so log lines /
+-- traceShow output match the on-the-wire JSON form instead of Latin-1 garbage.
+instance Show NameOwner where
+  show (NameOwner bs) = "NameOwner 0x" <> B.unpack (BAE.convertToBase BAE.Base16 bs)
```
epoberezkin: not needed?

src/Simplex/Messaging/Names/Owner.hs
```diff
@@ -0,0 +1,46 @@
...
+
+mkNameOwner :: ByteString -> Either String NameOwner
+mkNameOwner bs
+  | B.length bs == 20 = Right (NameOwner bs)
+  | otherwise = Left "NameOwner must be 20 bytes"
+
+unNameOwner :: NameOwner -> ByteString
+unNameOwner (NameOwner bs) = bs
```
epoberezkin: it's just selector in type

src/Simplex/Messaging/Names/Owner.hs
```diff
@@ -0,0 +1,46 @@
...
+import qualified Data.ByteArray.Encoding as BAE
+import Data.ByteString.Char8 (ByteString)
+import qualified Data.ByteString.Char8 as B
+import Data.Maybe (fromMaybe)
+import qualified Data.Text as T
+import Data.Text.Encoding (decodeLatin1, encodeUtf8)
+
+-- | 20-byte Ethereum address (NameRecord owner). Bare constructor not exported;
+-- use `mkNameOwner` to enforce the 20-byte invariant.
+newtype NameOwner = NameOwner ByteString
```
epoberezkin:
```suggestion
newtype NameOwner = NameOwner {unNameOwner :: ByteString}
```

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -796,6 +803,193 @@ validCountryValue field s
...
+--     are appended to it. Do not embed secrets in the path — it appears in
+--     logs; put credentials in resolver_auth.
+--   * on a non-loopback host, only http WITH resolver_auth is rejected (the
+--     Authorization header would travel in cleartext). http without auth is
+--     allowed (no secret to leak; resolver data is public — also lets dev
+--     setups reach a host resolver via host.docker.internal). https is always
+--     allowed, with or without auth.
+--   * link-local hosts (169.254.0.0/16, including the cloud metadata IP
+--     169.254.169.254) are rejected unconditionally
+validateUrl :: Text -> Maybe RpcAuth -> Either String Text
```
epoberezkin: why does it need to be so complex?

src/Simplex/Messaging/Agent.hs
```diff
@@ -1182,6 +1191,16 @@ getConnShortLink' c nm userId = \case
...
 deleteLocalInvShortLink' c (CSLInvitation _ srv linkId _) = withStore' c $ \db -> deleteInvShortLink db srv linkId
 
+resolveSimplexName' :: AgentClient -> NetworkRequestMode -> UserId -> SMPServer -> SimplexNameDomain -> AM NameRecord
+resolveSimplexName' c nm userId resolverSrv domain =
+  resolveName c nm userId resolverSrv placeholderContract (fullDomainName domain)
+  where
+    -- The wire format still carries a 20-byte `contract` field on RslvRequest
+    -- (no SMP version bump), but the server-side resolver ignores it: the
+    -- backing Python REST resolver is the source of truth for which on-chain
+    -- registry maps to each TLD. The agent sends the all-zero placeholder.
```
epoberezkin: Why pass placeholderContract here at all if it's not used?

epoberezkin: Shouldn't it be the actual contract we use for each TLD?

src/Simplex/Messaging/Protocol.hs
```diff
@@ -550,6 +564,19 @@ type LinkId = QueueId
...
+-- | Name resolution request. The client sends the name in canonical
+-- SimplexNameDomain form (TLD always explicit) as a Text plus the SNRC
+-- contract address it expects the server to query. The server parses the
+-- name into SimplexNameDomain (validating syntax) and checks the supplied
+-- contract against its hardcoded TLD whitelist before reading the chain —
+-- so a single names router can safely host multiple TLDs (each backed by
+-- its own SNRC contract) and reject clients that ask for the wrong one.
+data RslvRequest = RslvRequest
+  { name :: Text,
+    contract :: NameOwner
```
epoberezkin: NameOwner is a misnomer. It's suggests that it's the owner of the name, while it's used as an address of the contract. What I think is meant here is that it's _some_ ETH address, so the type should be named EthAddress, not NameOwner.

epoberezkin: also RslvRequest is unnecessary, it should be just fields in the command

src/Simplex/Messaging/Protocol.hs
```diff
@@ -597,6 +624,8 @@ data Command (p :: Party) where
   -- - entity ID: empty
   -- - corrId: unique correlation ID between proxy and relay, also used as a nonce to encrypt forwarded transmission
   RFWD :: EncFwdTransmission -> Command ProxyService -- use CorrId as CbNonce, proxy to relay
+  -- Name resolution: forwarded-only via PFWD. Server reads SNRC contract via Ethereum JSON-RPC.
+  RSLV :: RslvRequest -> Command Resolver
```
epoberezkin:
```suggestion
  RSLV :: EthAddress -> Text -> Command Resolver
```

src/Simplex/Messaging/Protocol.hs
```diff
@@ -705,6 +734,16 @@ instance Encoding FwdTransmission where
...
 
+instance J.ToJSON RslvRequest where
+  toJSON RslvRequest {name, contract} = J.object ["name" J..= name, "contract" J..= contract]
+  toEncoding RslvRequest {name, contract} = J.pairs ("name" J..= name <> "contract" J..= contract)
+
+instance J.FromJSON RslvRequest where
+  parseJSON = J.withObject "RslvRequest" $ \o -> do
+    name <- o J..: "name"
+    contract <- o J..: "contract"
+    pure RslvRequest {name, contract}
```
epoberezkin: not needed at all, and in general - don't hand-write instances, they must be derived from type

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1792,6 +1839,9 @@ instance PartyI p => ProtocolEncoding SMPVersion ErrorType (Command p) where
     PRXY host auth_ -> e (PRXY_, ' ', host, auth_)
     PFWD fwdV pubKey (EncTransmission s) -> e (PFWD_, ' ', fwdV, pubKey, Tail s)
     RFWD (EncFwdTransmission s) -> e (RFWD_, ' ', Tail s)
+    RSLV req
+      | v >= namesSMPVersion -> e (RSLV_, ' ', Tail (LB.toStrict (J.encode req)))
+      | otherwise -> e (ERR_, ' ', AUTH) -- pre-v20: shouldn't reach here, degrade to AUTH
```
epoberezkin: this is incorrect - the session version must be checked in the Client.hs, not in the encoder. You don't want to send ERR to the server, which is what it does.

Also, don't use JSON here, encode fields separately, as in other commands.

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1899,6 +1950,11 @@ instance ProtocolEncoding SMPVersion ErrorType Cmd where
     CT SNotifierService NSUBS_
       | v >= rcvServiceSMPVersion -> Cmd SNotifierService <$> (NSUBS <$> _smpP <*> smpP)
       | otherwise -> pure $ Cmd SNotifierService $ NSUBS (-1) mempty
+    CT SResolver RSLV_
+      | v >= namesSMPVersion -> do
+          Tail bs <- _smpP
+          either fail (pure . Cmd SResolver . RSLV) (J.eitherDecodeStrict bs)
```
epoberezkin: same here, no JSON needed

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1945,6 +2001,9 @@ instance ProtocolEncoding SMPVersion ErrorType BrokerMsg where
             | v < clientNoticesSMPVersion -> BLOCKED info {notice = Nothing}
           _ -> err
     PONG -> e PONG_
+    NAME rec
+      | v >= namesSMPVersion -> e (NAME_, ' ', Tail (LB.toStrict (J.encode rec)))
+      | otherwise -> e (ERR_, ' ', AUTH) -- pre-v20: shouldn't reach here, degrade to AUTH
```
epoberezkin: again, no need to check version here. If response arrived, it means it's supported. Version check is only needed when the same command may have different fields, not when it's just introduced.

src/Simplex/Messaging/Client.hs
```diff
@@ -1046,6 +1048,26 @@ sendSMPMessage c nm spKey sId flags msg =
 proxySMPMessage :: SMPClient -> NetworkRequestMode -> ProxiedRelay -> Maybe SndPrivateAuthKey -> SenderId -> MsgFlags -> MsgBody -> ExceptT SMPClientError IO (Either ProxyClientError ())
 proxySMPMessage c nm proxiedRelay spKey sId flags msg = proxyOKSMPCommand c nm proxiedRelay spKey sId (SEND flags msg)
 
+-- | Resolve a public-namespace name via PFWD. Preferred path - hides the
+-- client IP from the resolver. Mirrors `proxySMPMessage`'s shape; routes
+-- through `proxySMPCommand` and pattern-matches the expected NAME response.
+proxyResolveName :: SMPClient -> NetworkRequestMode -> ProxiedRelay -> NameOwner -> Text -> ExceptT SMPClientError IO (Either ProxyClientError NameRecord)
+proxyResolveName c nm proxiedRelay contract name =
```
epoberezkin: version check must exist here.

src/Simplex/Messaging/Client.hs
```diff
@@ -1046,6 +1048,26 @@ sendSMPMessage c nm spKey sId flags msg =
...
+    Right (NAME nr) -> pure $ Right nr
+    Right r -> throwE $ unexpectedResponse r
+    Left e -> pure $ Left e
+
+-- | Direct (non-PFWD) name resolution. Exposes the client IP to the resolver;
+-- callers that want anonymity should use `proxyResolveName` via the standard
+-- proxy fallback in the agent. RSLV requires no entity ID or authorization
+-- (see `noAuthCmd` in Protocol.hs).
+directResolveName :: SMPClient -> NetworkRequestMode -> NameOwner -> Text -> ExceptT SMPClientError IO NameRecord
+directResolveName c nm contract name =
```
epoberezkin: ... and here

src/Simplex/Messaging/Server/Names.hs
```diff
@@ -0,0 +1,162 @@
...
+
+data NamesConfig = NamesConfig
+  { resolverEndpoint :: Text,
+    resolverAuth :: Maybe RpcAuth,
+    resolverTimeoutMs :: Int,
+    resolverMaxResponseBytes :: Int
+  }
+  deriving (Show)
+
+data ResolveError
```
epoberezkin: You know we don't send this error to the clients, right, so we won't have any diagnostics visible to the users when something not working.

src/Simplex/Messaging/Server.hs
```diff
@@ -161,6 +163,13 @@ runSMPServer cfg attachHTTP_ = do
 runSMPServerBlocking :: MsgStoreClass s => TMVar Bool -> ServerConfig s -> Maybe AttachHTTP -> IO ()
 runSMPServerBlocking started cfg attachHTTP_ = newEnv cfg >>= runReaderT (smpServer started cfg attachHTTP_)
 
+-- | Test seam: run the server with a pre-built `namesEnv` (typically a stub
+-- backed by `newNamesEnvWith`). Production code MUST use `runSMPServerBlocking`,
+-- which builds `namesEnv` from `namesConfig` and probes the real RPC endpoint.
+runSMPServerBlockingWithNames :: MsgStoreClass s => TMVar Bool -> ServerConfig s -> Maybe AttachHTTP -> Maybe NamesEnv -> IO ()
+runSMPServerBlockingWithNames started cfg attachHTTP_ namesOverride =
```
epoberezkin: this is some antipattern. Whether we have names is already in config. So runSMPServerBlocking should already just work, this function is simply not needed.

src/Simplex/Messaging/Server/Env/STM.hs
```diff
@@ -558,7 +564,14 @@ newProhibitedSub = do
...
 newEnv :: ServerConfig s -> IO (Env s)
-newEnv config@ServerConfig {smpCredentials, httpCredentials, serverStoreCfg, smpAgentCfg, information, messageExpiration, idleQueueInterval, msgQueueQuota, maxJournalMsgCount, maxJournalStateLines} = do
+newEnv cfg = newEnvWithNames cfg Nothing
+
+-- | Test seam: build the server env, but if `namesOverride` is provided,
+-- use it as `namesEnv` and skip the production `newNamesEnv` / `pingEndpoint`
+-- path. This is the only injection point for stub `ethCall` implementations
+-- in functional-API tests.
+newEnvWithNames :: ServerConfig s -> Maybe NamesEnv -> IO (Env s)
+newEnvWithNames config@ServerConfig {allowSMPProxy, smpCredentials, httpCredentials, serverStoreCfg, smpAgentCfg, information, messageExpiration, idleQueueInterval, msgQueueQuota, maxJournalMsgCount, maxJournalStateLines, namesConfig} namesOverride = do
```
epoberezkin: Similarly this, as with server. Config already includes names - why a separate function is needed? newEnv should simply work with names

src/Simplex/Messaging/Server/Env/STM.hs
```diff
@@ -603,6 +616,22 @@ newEnv config@ServerConfig {smpCredentials, httpCredentials, serverStoreCfg, smp
   sockets <- newTVarIO []
   clientSeq <- newTVarIO 0
   proxyAgent <- newSMPProxyAgent smpAgentCfg random
+  namesEnv <- case namesOverride of
+    Just env -> pure (Just env)
+    Nothing -> case namesConfig of
+      Nothing -> pure Nothing
+      Just nc -> do
```
epoberezkin: this is messy and unnecessaryly complex

src/Simplex/Messaging/Server.hs
```diff
@@ -245,7 +254,10 @@ smpServer started cfg@ServerConfig {transports, transportConfig = tCfg, startOpt
       saveServerStats
 
     closeServer :: M s ()
-    closeServer = asks (smpAgent . proxyAgent) >>= liftIO . closeSMPClientAgent
+    closeServer = do
+      pa <- asks (smpAgent . proxyAgent)
+      ne <- asks namesEnv
+      liftIO $ closeSMPClientAgent pa `E.finally` mapM_ closeNamesEnv ne
```
epoberezkin: can be simpler and less diff

src/Simplex/Messaging/Server.hs
```diff
@@ -1486,6 +1505,22 @@ client
...
+        --   INTERNAL       - resolver / backing-store failure: transient, retry or surface
+        --   AUTH           - name not registered, or malformed name: authoritative "no such name"
+        (selector, msg) <- asks namesEnv >>= \case
+          Nothing -> pure (rslvDisabled, ERR $ CMD PROHIBITED)
+          Just nenv -> case parseName req of
+            Nothing -> pure (rslvBadName, ERR AUTH)
+            Just d -> liftIO (resolveName nenv d) <&> \case
+              Right rec -> (rslvSucc, NAME rec)
+              Left NotFound -> (rslvNotFound, ERR AUTH)
+              Left _ -> (rslvEthErrs, ERR INTERNAL)
```
epoberezkin: This is messy. First, errors would be confusing in the client. Second, we mix parsing and processing. There is an argument to fail invalid names at parsing stage. Alternatively, we need some new error vocabulary in ErrorType maybe? Not sure.

src/Simplex/Messaging/Names/Owner.hs
epoberezkin: Should be EthAddress

src/Simplex/Messaging/Names/Record.hs
```diff
@@ -0,0 +1,89 @@
...
+
+-- | Resolved name record returned by the names role.
+--   Wire format is JSON — change requires an SMP version bump.
+--   JSON keys match the Python REST resolver (`snrc-resolve.py`).
+--   Text fields use the empty string as the "unset" sentinel; coin fields
+--   use JSON `null`. simplexContact / simplexChannel are arrays of links
+--   (primary first, empty array when unset) so a name can advertise fallback
+--   SMP servers. `owner` and `resolver` carry 20-byte addresses encoded as
+--   `0x`-prefixed lowercase hex (see Names.Owner).
+data NameRecord = NameRecord
```
epoberezkin, line 25: should match JSON

src/Simplex/Messaging/Server/Main.hs
epoberezkin: and why does it not use Uri type from the library we have?

src/Simplex/Messaging/Agent.hs
epoberezkin: My understanding was that we will be sending REAL contract addresses that server knows, and we would be validating them on the server side. This would allow extending and switching. Currently, the address client sends is just ignored (and client sends dummy address).

What's the idea here? Postpone it till we actuall have another address and simply reserve the name in the syntax? why not send real address, at least? Because if not, we'd later have to interpret dummy address server-side as meaning the current address.

tests/SMPClient.hs
```diff
@@ -362,6 +364,16 @@ withSmpServerConfigOn t (ASrvCfg _ _ cfg') port' =
     (\started -> runSMPServerBlocking started cfg' {transports = [(port', t, False)]} Nothing)
     (threadDelay 10000)
 
+-- | Variant of `withSmpServerConfigOn` for RSLV functional tests: passes a
+-- pre-built `NamesEnv` (typically with a stub `ethCall`) so the server does
+-- not contact the real Ethereum RPC. Skips the production `pingEndpoint`
+-- probe.
+withSmpServerConfigOnWithNames :: HasCallStack => ASrvTransport -> AServerConfig -> ServiceName -> NamesEnv -> (HasCallStack => ThreadId -> IO a) -> IO a
+withSmpServerConfigOnWithNames t (ASrvCfg _ _ cfg') port' nenv =
```
epoberezkin: this function is also not needed - this must be just custom config including names

src/Simplex/Messaging/Server.hs
epoberezkin: especially if it's used only in tests it must be removed.

epoberezkin (changes requested):

src/Simplex/Messaging/Agent.hs
```diff
@@ -440,6 +442,13 @@ getConnShortLink :: AgentClient -> NetworkRequestMode -> UserId -> ConnShortLink
 getConnShortLink c = withAgentEnv c .:. getConnShortLink' c
 {-# INLINE getConnShortLink #-}
 
+-- | Resolve a SimpleX name via the configured resolver SMP server (PFWD RSLV).
+-- The TLD->contract whitelist lives in the agent so chat clients only need to
+-- pass the resolver address and the parsed domain.
+resolveSimplexName :: AgentClient -> NetworkRequestMode -> UserId -> SMPServer -> SimplexNameDomain -> AE NameRecord
```
epoberezkin:
```suggestion
resolveSimplexName :: AgentClient -> NetworkRequestMode -> UserId -> SimplexNameDomain -> AE NameRecord
```

Server should be chosen randomly from supported servers based on ServerRole where names::Bool must be added

src/Simplex/Messaging/Names/Record.hs
```diff
@@ -0,0 +1,63 @@
...
+-- the resolver output for unset coins).
+$( JQ.deriveJSON
+    defaultJSON {J.omitNothingFields = False, J.fieldLabelModifier = dropPrefix "nr"}
+    ''NameRecord
+ )
+
+-- Wire encoding for the SMP NAME response: field-ordered smpEncode, not embedded
+-- JSON. Field order = record declaration order. EthAddress encodes as its raw
+-- 20 bytes (length-prefixed via the ByteString instance).
+instance Encoding NameRecord where
```
epoberezkin: remove

epoberezkin: wire encoding should use JSON

src/Simplex/Messaging/Server/Env/STM.hs
```diff
@@ -603,6 +611,23 @@ newEnv config@ServerConfig {smpCredentials, httpCredentials, serverStoreCfg, smp
   sockets <- newTVarIO []
   clientSeq <- newTVarIO 0
   proxyAgent <- newSMPProxyAgent smpAgentCfg random
+  namesEnv <- case namesConfig of
+    Nothing -> pure Nothing
+    Just nc -> case namesResolverCall_ of
```
evgeny-simplex:
```suggestion
  namesEnv <- forM namesConfig $ \nc -> case namesResolverCall_ of
```

src/Simplex/Messaging/Server/Env/STM.hs
```diff
@@ -603,6 +611,23 @@ newEnv config@ServerConfig {smpCredentials, httpCredentials, serverStoreCfg, smp
...
+      -- test seam: stub resolver, no real HTTP env or startup probe
+      Just call -> Just <$> newNamesEnvWith nc call Nothing
```
evgeny-simplex:
```suggestion
      Just call -> newNamesEnvWith nc call Nothing
```

src/Simplex/Messaging/Server/Env/STM.hs
```diff
@@ -603,6 +611,23 @@ newEnv config@ServerConfig {smpCredentials, httpCredentials, serverStoreCfg, smp
...
+        when allowSMPProxy $
+          logWarn "[NAMES] enable: on on a proxy-role host: slow RSLV calls can serialise other forwarded commands on the same proxy-relay session. For high-volume deployments, run [NAMES] on a separate host."
+        env <- newNamesEnv nc
+        -- Probe the endpoint at startup. Don't exitFailure: a flapping
+        -- network or an Ethereum host coming up minutes after smp-server
+        -- should not block the server. Log so operators can spot it.
+        pingEndpoint env >>= \case
+          Right _ -> logInfo "[NAMES] endpoint probe ok"
+          Left e -> logWarn $ "[NAMES] endpoint probe failed (server will still start, RSLV will return ERR (NAME ...) until reachable): " <> tshow e
+        pure (Just env)
```
evgeny-simplex:
```suggestion
        pure env
```

src/Simplex/Messaging/Server/Env/STM.hs
```diff
@@ -603,6 +611,23 @@ newEnv config@ServerConfig {smpCredentials, httpCredentials, serverStoreCfg, smp
...
   proxyAgent <- newSMPProxyAgent smpAgentCfg random
+  namesEnv <- case namesConfig of
+    Nothing -> pure Nothing
+    Just nc -> case namesResolverCall_ of
+      -- test seam: stub resolver, no real HTTP env or startup probe
+      Just call -> Just <$> newNamesEnvWith nc call Nothing
+      Nothing -> do
+        logInfo $ "[NAMES] resolver enabled, endpoint=" <> scrubUrl (resolverEndpoint nc)
+        when allowSMPProxy $
+          logWarn "[NAMES] enable: on on a proxy-role host: slow RSLV calls can serialise other forwarded commands on the same proxy-relay session. For high-volume deployments, run [NAMES] on a separate host."
```
evgeny-simplex: why? this is a strange comment. We allow SMP proxy on all hosts, name resolution should be async in the same way all proxy calls are async.

src/Simplex/Messaging/Server/Env/STM.hs
evgeny-simplex: this is a strange way to inject stub. Shouldn't it be the same endpoint?

src/Simplex/Messaging/Server/Main/Init.hs
```diff
@@ -155,6 +155,22 @@ iniFileContent cfgPath logPath opts host basicAuth controlPortPwds =
...
     <> ("# client_concurrency = " <> tshow defaultProxyClientConcurrency)
     <> "\n\n\
+        \[NAMES]\n\
+        \# Public-namespace resolution via the snrc-resolve.py REST resolver.\n\
+        \# Operator runs the resolver alongside smp-server (default port 8000)\n\
+        \# with its own Ethereum JSON-RPC endpoint configured in resolver.toml.\n\
+        \# Co-locating with the proxy role logs a startup advisory: slow RSLV calls can\n\
+        \# serialise other forwarded commands on the same proxy-relay session.\n\
+        \# For high-volume deployments, run [NAMES] on a separate host.\n\
+        \# Restart required to change settings.\n\
```
evgeny-simplex: this is wrong and should not be the case

evgeny-simplex: also, we don't colocate it with SMP, right?

src/Simplex/Messaging/Server/Names/HttpResolver.hs
```diff
@@ -0,0 +1,164 @@
...
+    let status = HT.statusCode (responseStatus res)
+    if status >= 400
+      then pure (Left (HttpStatusErr status))
+      else do
+        bs <- brReadSome (responseBody res) (maxResponseBytes + 1)
+        if BL.length bs > fromIntegral maxResponseBytes
+          then pure (Left BodyTooLarge)
+          else case J.eitherDecodeStrict (BL.toStrict bs) of
+            Left e -> pure (Left (InvalidJson e))
+            Right v -> pure (Right v)
```
evgeny-simplex:
```suggestion
          else first InvalidJson <$> J.eitherDecodeStrict (BL.toStrict bs)
```

evgeny-simplex: in general, please review and refactor such trivial cases into forM, mapM, first, second, bimap, maybe, etc.

src/Simplex/Messaging/Server/Names/HttpResolver.hs
```diff
@@ -0,0 +1,164 @@
...
+resolveHttp :: ResolverEnv -> Text -> IO (Either ResolverError J.Value)
+resolveHttp env name = doGet env ("/resolve/" <> percentEncode name)
+
+-- | GET <baseUrl>/health, return the JSON body on 200.
+healthHttp :: ResolverEnv -> IO (Either ResolverError J.Value)
+healthHttp env = doGet env "/health"
+
+doGet :: ResolverEnv -> Text -> IO (Either ResolverError J.Value)
+doGet ResolverEnv {manager, baseUrl, authHdr, timeoutMicro, maxResponseBytes} path = do
+  req0 <- parseRequest (T.unpack (baseUrl <> path))
```
evgeny-simplex: this T.unpack is a smell that incorrect type is passed from upstream

src/Simplex/Messaging/Server/Names/HttpResolver.hs
```diff
@@ -0,0 +1,164 @@
...
+            Left e -> pure (Left (InvalidJson e))
+            Right v -> pure (Right v)
+  pure (either (Left . HttpFailure) id result)
+
+-- | Percent-encode a name component (path-safe). Aggressive: encode every
+-- byte that isn't an unreserved character per RFC 3986. The resolver expects
+-- raw labels (e.g., `alice.simplex`); slashes and other ASCII punctuation
+-- would change the request path semantics if passed through verbatim.
+percentEncode :: Text -> Text
+percentEncode = decodeLatin1 . urlEncode True . encodeUtf8
```
evgeny-simplex: trivial function that is not needed if you use correct types (String or ByteString)

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -796,6 +804,198 @@ validCountryValue field s
...
+            NamesConfig
+              { resolverEndpoint = either (error . ("[NAMES] resolver_endpoint: " <>)) id (validateUrl endpoint resolverAuth_),
+                resolverAuth = resolverAuth_,
+                resolverTimeoutMs = boundedIniInt 3000 100 60000 "resolver_timeout_ms",
+                -- ceiling = SMP transport budget: the NAME response is one SMP
+                -- transmission (proxied: padded to paddedProxiedTLength = 16226),
+                -- and the smpEncoded NameRecord is <= its JSON body, so capping
+                -- the body here guarantees the response always frames. An
+                -- over-cap body fails as BodyTooLarge -> ERR (NAME (RESOLVER ..)).
+                resolverMaxResponseBytes = boundedIniInt 16000 1024 16000 "resolver_max_response_bytes"
```
evgeny-simplex, line 815: and in some other place it's 64kb and no 16. If we need larger size, we could use zstd compression in SMP protocol.

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -796,6 +804,198 @@ validCountryValue field s
...
+  unless (null (uriUserInfo ua)) $ Left "userinfo (user:pass@) not allowed; use resolver_auth instead"
+  case uriPort ua of
+    "" -> Left "explicit port required (e.g. http://host:8000)"
+    ':' : portStr -> case readMaybe portStr of
+      Just n | (n :: Int) >= 1 && n <= 65535 -> Right ()
+      _ -> Left $ "port " <> portStr <> " out of range (must be 1..65535)"
+    other -> Left $ "unexpected port syntax: " <> other
+  unless (null (uriQuery uri)) $ Left "query string not allowed (it does not compose with the appended /resolve/<name> path)"
+  unless (null (uriFragment uri)) $ Left "fragment not allowed (fragments are never sent to the server)"
+  -- A path prefix is allowed and used as the base for /resolve/<name> and
```
evgeny-simplex: this is very complex and unreadable, needs to be simplified

evgeny-simplex: we use library for URI parsing, why we are re-implement it here?

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -796,6 +804,198 @@ validCountryValue field s
...
+--     logs; put credentials in resolver_auth.
+--   * on a non-loopback host, only http WITH resolver_auth is rejected (the
+--     Authorization header would travel in cleartext). http without auth is
+--     allowed (no secret to leak; resolver data is public — also lets dev
+--     setups reach a host resolver via host.docker.internal). https is always
+--     allowed, with or without auth.
+--   * link-local hosts (169.254.0.0/16, including the cloud metadata IP
+--     169.254.169.254) are rejected unconditionally
+validateUrl :: Text -> Maybe RpcAuth -> Either String Text
+validateUrl url auth_ = do
```
evgeny-simplex, line 835: this function needs to be fully re-written or used from library. It has to be like 1-3 lines of code, not 50.

src/Simplex/Messaging/Server/Main.hs
```diff
@@ -796,6 +804,198 @@ validCountryValue field s
...
+        high96Zero = w1 == 0 && w2 == 0
+        compatTo169 = high96Zero && w3 == 0 && is169254v4
+        mappedTo169 = high96Zero && w3 == 0xffff && is169254v4
+        sixToFour169 = (w1 `shiftR` 16) == 0x2002 && (w1 .&. 0xffff) == 0xa9fe
+        nat64To169 = w1 == 0x0064ff9b && w2 == 0 && w3 == 0 && is169254v4
+
+-- | Parse an rpc_auth INI value. Scheme keyword is case-insensitive so
+-- "Bearer <token>" / "BEARER <token>" (Caddy / RFC 7235 convention) work
+-- as well as the lowercase form.
+parseRpcAuth :: Text -> Either String RpcAuth
```
evgeny-simplex, line 853: same here - it's an existing library function, it should not be in our code.

evgeny-simplex: the library is uri-bytestring, just needs to be added and used

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1565,10 +1588,27 @@ data ErrorType
...
+-- One vocabulary shared server-side and agent-side so name failures flow
+-- through the single error type to chat (as ChatErrorAgent) with diagnostics,
+-- mirroring ProxyError.
+data NameErrorType
+  = -- | the names role / resolver is not configured on this server
+    NO_RESOLVER
+  | -- | the name is not registered (resolver returned not-found)
+    NO_NAME
+  | -- | no name-resolving servers configured (agent-originated only)
+    NO_SERVERS
```
evgeny-simplex: as it's agent-originated, it must not be in this type - it must be in Agent errors

evgeny-simplex: This type is only for errors returned by the server, it can't contain errors reported upstream

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1565,10 +1588,27 @@ data ErrorType
...
 
+-- | Name resolution errors (the NAME family of ErrorType / AgentErrorType).
+-- One vocabulary shared server-side and agent-side so name failures flow
+-- through the single error type to chat (as ChatErrorAgent) with diagnostics,
+-- mirroring ProxyError.
+data NameErrorType
+  = -- | the names role / resolver is not configured on this server
+    NO_RESOLVER
+  | -- | the name is not registered (resolver returned not-found)
+    NO_NAME
```
evgeny-simplex: NOT_FOUND would be more typical

There may also be expired error - I don't know how it will be reported by resolver, so probably not needed for now.

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1899,6 +1944,11 @@ instance ProtocolEncoding SMPVersion ErrorType Cmd where
     CT SNotifierService NSUBS_
       | v >= rcvServiceSMPVersion -> Cmd SNotifierService <$> (NSUBS <$> _smpP <*> smpP)
       | otherwise -> pure $ Cmd SNotifierService $ NSUBS (-1) mempty
+    -- Name is validated at parse (invalid syntax fails here -> CMD error),
+    -- so the handler only ever sees a valid SimplexNameDomain.
```
evgeny-simplex:
```suggestion
```

redundant

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1792,6 +1834,8 @@ instance PartyI p => ProtocolEncoding SMPVersion ErrorType (Command p) where
     PRXY host auth_ -> e (PRXY_, ' ', host, auth_)
     PFWD fwdV pubKey (EncTransmission s) -> e (PFWD_, ' ', fwdV, pubKey, Tail s)
     RFWD (EncFwdTransmission s) -> e (RFWD_, ' ', Tail s)
+    -- Version gating is the client's job (Client.hs), not the encoder's.
```
evgeny-simplex:
```suggestion
```

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1792,6 +1834,8 @@ instance PartyI p => ProtocolEncoding SMPVersion ErrorType (Command p) where
...
+    RSLV d -> e (RSLV_, ' ', Tail (strEncode d))
```
evgeny-simplex:
```suggestion
    RSLV d -> e (RSLV_, ' ', d)
```

evgeny-simplex: maybe I am missing something, but using Tail explicitly here is redundant - all it achieves is preventing length prefixing of the string. Directly using d here achieves the same.

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1899,6 +1944,11 @@ instance ProtocolEncoding SMPVersion ErrorType Cmd where
     CT SNotifierService NSUBS_
       | v >= rcvServiceSMPVersion -> Cmd SNotifierService <$> (NSUBS <$> _smpP <*> smpP)
       | otherwise -> pure $ Cmd SNotifierService $ NSUBS (-1) mempty
+    -- Name is validated at parse (invalid syntax fails here -> CMD error),
+    -- so the handler only ever sees a valid SimplexNameDomain.
+    CT SResolver RSLV_ -> do
+      Tail bs <- _smpP
+      either fail (pure . Cmd SResolver . RSLV) (strDecode bs)
```
evgeny-simplex:
```suggestion
      Cmd . SResolver . RSLV <$> _smpP
```

evgeny-simplex: same here

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1945,6 +1995,9 @@ instance ProtocolEncoding SMPVersion ErrorType BrokerMsg where
             | v < clientNoticesSMPVersion -> BLOCKED info {notice = Nothing}
           _ -> err
     PONG -> e PONG_
+    -- Field-ordered Encoding NameRecord (no JSON on the wire); a response that
+    -- arrived is already on a supported version, so no version gate.
+    RNAME rec -> e (RNAME_, ' ', rec)
```
evgeny-simplex:
```suggestion
    RNAME rec -> e (RNAME_, ' ', Tail $ toStrict $ J.encode rec)
```

src/Simplex/Messaging/Protocol.hs
```diff
@@ -1992,6 +2045,7 @@ instance ProtocolEncoding SMPVersion ErrorType BrokerMsg where
     OK_ -> pure OK
     ERR_ -> ERR <$> _smpP
     PONG_ -> pure PONG
+    RNAME_ -> RNAME <$> _smpP
```
evgeny-simplex:
```suggestion
    RNAME_ -> fmap RNAME . J.decodeStrict . unTail <$?> _smpP
```

src/Simplex/Messaging/Protocol.hs
```diff
@@ -2074,9 +2130,28 @@ instance Encoding ErrorType where
...
       "INTERNAL" -> pure INTERNAL
+      "NAME" -> NAME <$> _smpP
       "DUPLICATE_" -> pure DUPLICATE_
       _ -> fail "bad ErrorType"
 
+instance Encoding NameErrorType where
+  smpEncode = \case
+    NO_RESOLVER -> "NO_RESOLVER"
+    NO_NAME -> "NO_NAME"
+    NO_SERVERS -> "NO_SERVERS"
```
evgeny-simplex: see above

src/Simplex/Messaging/Server.hs
```diff
@@ -1486,6 +1497,19 @@ client
...
       Cmd SProxyService (RFWD encBlock) -> response . (corrId,NoEntity,) <$> processForwardedCommand encBlock
+      Cmd SResolver (RSLV d) -> do
+        st <- asks (rslvStats . serverStats)
+        incStat (rslvReqs st)
+        -- The name is validated at command parse (invalid syntax never reaches
+        -- here), so the handler only maps the resolver outcome to a declared
+        -- error that reaches the client as ERR (NAME ...).
+        (selector, msg) <- asks namesEnv >>= \case
+          Nothing -> pure (rslvDisabled, ERR $ NAME NO_RESOLVER)
+          Just nenv -> liftIO (resolveName nenv d) <&> \case
```
evgeny-simplex: that's the crux of the problem. You call resolver inline, blocking further command processing (and that's what all those comments were about). This has to use the same pattern as we use to create proxy session - it must be processed asynchronously, and response send from forked thread. The function simply returns Nothing as command response, deferring response to forked thread.

src/Simplex/Messaging/Server.hs
```diff
@@ -1262,6 +1270,9 @@ verifyQueueTransmission service thAuth (tAuth, authorized, (corrId, entId, comma
     vc SNotifierService NSUBS {} = verifyServiceCmd
     vc SProxiedClient _ = VRVerified Nothing
     vc SProxyService (RFWD _) = VRVerified Nothing
+    -- RSLV is accepted both forwarded (via PFWD, preferred - hides client IP from resolver)
+    -- and direct (client->resolver, faster, exposes client IP). Mode is chosen by the client.
```
epoberezkin:
```suggestion
```

src/Simplex/Messaging/Server.hs
```diff
@@ -2083,45 +2141,62 @@ client
...
-          r2 <- liftEitherWith (const BLOCK) $ EncResponse <$> C.cbEncrypt clientSecret (C.reverseNonce clientNonce) r' paddedProxiedTLength
-          -- encrypt to proxy
-          let fr = FwdResponse {fwdCorrId, fwdResponse = r2}
-              r3 = EncFwdResponse $ C.cbEncryptNoPad sessSecret (C.reverseNonce proxyNonce) (smpEncode fr)
-          stats <- asks serverStats
-          incStat $ pMsgFwdsRecv stats
-          pure r3
+        -- Returns Nothing for a forwarded RSLV: like proxying, it resolves and
+        -- replies from a forked thread (forkCmd writes the RRES to sndQ), so a
+        -- slow RSLV does not serialise other forwarded commands on this session.
```
epoberezkin:
```suggestion
```

PR #1788, opened 2026-05-28
agent: split SimplexNameDomain out of SimplexNameInfo

shumvgolove:


src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -1540,40 +1548,47 @@ instance StrEncoding SimplexNameType where
...
-          | nt == NTPublicGroup -> Right $ SimplexNameInfo nt TLDSimplex name []
-          | otherwise -> Left "contact name requires TLD"
-        tld : name : sub -> Right $ case tld of
-          "simplex" -> SimplexNameInfo nt TLDSimplex name sub
-          "testing" -> SimplexNameInfo nt TLDTesting name sub
-          _ -> SimplexNameInfo nt TLDWeb (T.intercalate "." labels) []
-
-fullDomainName :: SimplexNameInfo -> Text
-fullDomainName SimplexNameInfo {nameTLD, domain, subDomain} = T.intercalate "." (reverse subDomain ++ [domain] ++ tld')
+        [name] -> Right $ SimplexNameDomain TLDSimplex name []
```
epoberezkin: this is not a valid domain, it's only valid as part of channel name

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -1540,40 +1548,47 @@ instance StrEncoding SimplexNameType where
...
+        "simplex" : name : sub -> Right $ SimplexNameDomain TLDSimplex name sub
+        "testing" : name : sub -> Right $ SimplexNameDomain TLDTesting name sub
+        _ -> Right $ SimplexNameDomain TLDWeb (T.intercalate "." labels) []
+
+instance StrEncoding SimplexNameInfo where
```
epoberezkin: move above

src/Simplex/Messaging/Agent/Protocol.hs
```diff
@@ -1540,40 +1548,47 @@ instance StrEncoding SimplexNameType where
...
+  strEncode SimplexNameInfo {nameType, nameDomain} =
+    "simplex:/name" <> strEncode nameType <> strEncode nameDomain
+  strP = optional "simplex:/name" *> ((strP >>= infoP) <|> infoP NTPublicGroup)
+    where
+      infoP NTContact = do
+        bs <- lookAhead $ A.takeWhile1 (not . A.isSpace)
+        unless (B.elem '.' bs) $ fail "contact name requires TLD"
```
epoberezkin: it should be the opposite. Not contact is more restrictive, but channel should be more permissive than domain parser

src/Simplex/Messaging/Agent/Protocol.hs
hakanferi412-sys: #729

PR #1801, opened 2026-06-05
core: validate SNTRUP761 KEM input lengths

pbottine:
## Summary
- Validate SNTRUP761 public key, secret key, and ciphertext lengths when parsing encoded values.
- Reject malformed public keys and ciphertexts before calling the C KEM bindings.
- Add regression tests for malformed encodings and direct FFI wrapper inputs.

## Tests
- `cabal test simplexmq-test --extra-include-dirs=/opt/homebrew/opt/openssl@3/include --extra-lib-dirs=/opt/homebrew/opt/openssl@3/lib --test-option=--match --test-option='/Encryption tests/sntrup761/'`


epoberezkin:
fixed in #1811

PR #1802, opened 2026-06-08
smp: reject duplicate STM short link updates

pbottine:
## Summary
- Reject STM `LSET` attempts that claim a `LinkId` already mapped to a different recipient queue.
- Keep same-queue link updates working.
- Add a regression that creates a victim short link, attempts to attach that link id to a separate authenticated queue, and verifies the victim link still resolves to the victim data.

## Tests
- `cabal test simplexmq-test --extra-include-dirs=/opt/homebrew/opt/openssl@3/include --extra-lib-dirs=/opt/homebrew/opt/openssl@3/lib --test-option=--match --test-option='rejects attaching an existing short link to another queue'`


epoberezkin:
moved to #1802

PR #1803, opened 2026-06-08
xftp: count only committed files as used storage

pbottine:
## Summary
- Count only XFTP files with a committed `file_path` as used storage.
- Apply the same accounting rule to STM and Postgres file stores.
- Keep pending rows visible in file count accounting.
- Add store tests covering pending and committed storage totals.

## Tests
- `PGHOST=127.0.0.1 cabal test simplexmq-test -fserver_postgres --extra-include-dirs=/opt/homebrew/opt/openssl@3/include --extra-lib-dirs=/opt/homebrew/opt/openssl@3/lib --test-option=--match --test-option='committed used storage and file count'`


epoberezkin:
moved to #1814

PR #1807, opened 2026-06-15
lib: parse bracketed IPv6 server addresses

shumvgolove:
ref: https://github.com/trailofbits/ptp-simplexmq/issues/9

### DIFF
--- a/src/Simplex/Messaging/ServiceScheme.hs
+++ b/src/Simplex/Messaging/ServiceScheme.hs
@@ -11,8 +11,9 @@ import Control.Applicative ((<|>))
 import qualified Data.Attoparsec.ByteString.Char8 as A
 import qualified Data.ByteString.Char8 as B
 import Data.Functor (($>))
-import Network.Socket (HostName, ServiceName)
+import Network.Socket (ServiceName)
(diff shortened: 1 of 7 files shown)


src/Simplex/Messaging/ServiceScheme.hs
```diff
@@ -25,14 +26,13 @@ instance StrEncoding ServiceScheme where
...
       <|> "https://" *> (SSAppServer <$> strP)
 
-data SrvLoc = SrvLoc HostName ServiceName
+data SrvLoc = SrvLoc TransportHost ServiceName
   deriving (Eq, Ord, Show)
 
 instance StrEncoding SrvLoc where
-  strEncode (SrvLoc host port) = B.pack $ host <> if null port then "" else ':' : port
-  strP = SrvLoc <$> host <*> (port <|> pure "")
+  strEncode (SrvLoc host port) = strEncode host <> B.pack (if null port then "" else ':' : port)
```
epoberezkin:
```suggestion
  strEncode (SrvLoc host port) = case host of
     THIPv6 _ | not (null port) -> strEncode ('[',  host, ']') <> B.pack (':' : port)
     _ -> strEncode host
```

PR #1809, opened 2026-06-18
core: add getentropy shim for windows build

shumvgolove:


simplexmq.cabal
```diff
@@ -333,6 +334,9 @@ library
       cbits/blst/build/assembly.S
   extra-libraries:
       crypto
+  if os(windows)
```
epoberezkin, line 336:
```suggestion
  if os(windows)
    c-sources:
        cbits/getentropy_win.c
```

simplexmq.cabal
```diff
@@ -323,6 +323,7 @@ library
       cbits/sha512.c
       cbits/sntrup761.c
       cbits/blst/src/server.c
+      cbits/getentropy_win.c
```
epoberezkin:
```suggestion
```

PR #1825, opened 2026-07-06
smp-server: add serverInfoBytes to THandleParams and related functions

ed-asriyan:


src/Simplex/Messaging/Transport.hs
```diff
@@ -496,7 +501,9 @@ data THandleParams v p = THandleParams
     -- based on protocol version
     batch :: Bool,
     -- | include service signature (or '0' if it is absent), based on protocol version
-    serviceAuth :: Bool
+    serviceAuth :: Bool,
+    -- | JSON-encoded ServerPublicInfo from handshake, present when server version >= serverInfoSMPVersion
+    serverInfoBytes :: Maybe ByteString
```
epoberezkin, line 506: why not include the actual type here, and parse JSON in parser? Or you think there is a value in lenient parser? but then you could just include Nothing if JSON fails to parse, or Either String ServerInfo (it would help to know in test that info was sent but was invalid).

```suggestion
    serverInfo :: Either String ServerPublicInfo
```

src/Simplex/Messaging/Transport.hs
```diff
@@ -548,7 +555,9 @@ data SMPServerHandshake = SMPServerHandshake
     sessionId :: SessionId,
     -- pub key to agree shared secrets for command authorization and entity ID encryption.
     -- todo C.PublicKeyX25519
-    authPubKey :: Maybe CertChainPubKey
+    authPubKey :: Maybe CertChainPubKey,
+    -- | optional server public information (JSON-encoded ServerPublicInfo), sent when version >= serverInfoSMPVersion
+    serverInfo :: Maybe ByteString
```
epoberezkin, line 560:
```suggestion
    serverInfo :: Either String ServerPublicInfo
```

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -205,7 +205,9 @@ import Control.Monad.Reader
 import Control.Monad.Trans.Except
 import Crypto.Random (ChaChaDRG)
 import qualified Data.Aeson as J
+import qualified Data.Aeson as Aeson
```
epoberezkin:
```suggestion
```

Already imported above as J, but it'll move to transport if we put it into handshake

src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -1281,7 +1284,7 @@ data ProtocolTestFailure = ProtocolTestFailure
   }
   deriving (Eq, Show)
 
-runSMPServerTest :: AgentClient -> NetworkRequestMode -> UserId -> SMPServerWithAuth -> AM' (Maybe ProtocolTestFailure)
+runSMPServerTest :: AgentClient -> NetworkRequestMode -> UserId -> SMPServerWithAuth -> AM' (Either ProtocolTestFailure (Maybe ServerPublicInfo))
```
epoberezkin:
```suggestion
runSMPServerTest :: AgentClient -> NetworkRequestMode -> UserId -> SMPServerWithAuth -> AM' (Either ProtocolTestFailure (Either String ServerPublicInfo))
```

It would help differentiate info not sent/failed to parse scenarios, and this type would already be included into handshake and handle, so all parsing would be in one place, and not split between transport and client.

src/Simplex/Messaging/Transport.hs
ed-asriyan: I tried replacing `Maybe ByteString` with `Either String ServerPublicInfo`, but putting it into the Transport.hs exports creates a cyclic dependency `Simplex.Messaging.Transport` -> `Simplex.Messaging.Server.Information` -> `Simplex.Messaging.Agent.Protocol` -> `Simplex.Messaging.Transport.Client` -> `Simplex.Messaging.Transport`.

I'm not export in Haskell - it's literally the first time for me writing on Haskell :) - but I guess untangling the dependencies may require more refactoring in other modules. Do you think it may worth it?

PR #1839, opened 2026-07-30
smp server: do not create messaging queues in SMP proxy to prevent deadlock in processing

shumvgolove:


src/Simplex/Messaging/Client/Agent.hs
```diff
@@ -162,7 +162,9 @@ newSMPClientAgent :: SParty p -> SMPClientAgentConfig -> Maybe DBService -> TVar
 newSMPClientAgent agentParty agentCfg@SMPClientAgentConfig {msgQSize, agentQSize} dbService randomDrg = do
   active <- newTVarIO True
   startedAt <- getCurrentTime
-  msgQ <- newTBQueueIO msgQSize
+  -- Only subscribing agents receive server transmissions. A queue nobody reads fills with late
+  -- responses (Client.hs, processMsg) until the client's `process` thread blocks writing to it.
```
epoberezkin:
```suggestion
  -- Only subscribing agents receive server transmissions, should not be created until processed to prevent deadlock
```

src/Simplex/Messaging/Client/Agent.hs
```diff
@@ -162,7 +162,9 @@ newSMPClientAgent :: SParty p -> SMPClientAgentConfig -> Maybe DBService -> TVar
...
+  msgQ <- forM (serviceParty agentParty) $ \_ -> newTBQueueIO msgQSize
```
epoberezkin: It conflates the idea that agent can be a service with agent will process messages - it accidentally coincides, but semantically is different.

A better approach is to replace msgQSize :: Natural with msgQSize :: Maybe Natural in config

PR #1842, opened 2026-08-05
agent: report PRXY broker errors as proxy-to-relay errors

Narasimha-sc:
With private routing enabled, a failure to reach the destination relay is reported as "Error connecting to forwarding server <preset server>", blaming a server the client connected to successfully; retrying just names the next preset server.

The proxy returns `PROXY (BROKER ...)` only when it fails to reach the relay, but `PRXY` responses were mapped to `SMP <proxy> (PROXY ...)` (`Agent/Client.hs:713`), discarding the relay address, so both clients read them as client-to-proxy failures.

This maps that one class to `PROXY {proxyServer, relayServer, ...}` — the shape `PFWD` errors already use — so the existing client alerts name both servers with no app change. Rationale, guard width and blast radius in `plans/2026-08-05-proxy-relay-error-attribution.md`.

### DIFF
--- a/src/Simplex/Messaging/Agent/Client.hs
+++ b/src/Simplex/Messaging/Agent/Client.hs
@@ -710,7 +710,7 @@ getSMPProxyClient c@AgentClient {active, smpClients, smpProxiedRelays, workerSeq
       pure (clnt, sess)
     newProxiedRelay :: SMPConnectedClient -> Maybe SMP.BasicAuth -> ProxiedRelayVar -> AM (Either AgentErrorType ProxiedRelay)
     newProxiedRelay (SMPConnectedClient smp prs) proxyAuth rv =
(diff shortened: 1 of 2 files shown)


src/Simplex/Messaging/Agent/Client.hs
```diff
@@ -725,6 +721,18 @@ getSMPProxyClient c@AgentClient {active, smpClients, smpProxiedRelays, workerSeq
...
+      where
+        -- proxy reports BROKER errors about the relay, not about its own connection,
+        -- so they include both addresses, same as PFWD errors.
+        proxyRelayError :: HostName -> ErrorType -> AgentErrorType
+        proxyRelayError proxyHost = \case
+          e@(SMP.PROXY (SMP.BROKER _)) ->
+            PROXY
+              { proxyServer = protocolClientServer smp,
+                relayServer = B.unpack $ strEncode destSrv,
+                proxyErr = ProxyProtocolError e
```
epoberezkin:
```suggestion
                proxyErr = ProxyResponseError e
```

PR #1846, opened 2026-08-09
feat: add server public information handling in XFTP protocol

ed-asriyan:


ed-asriyan:
The PR should be merged when most of the users will be upgraded to the version that ignores trailing bytes in the handshake (see #1852)

PR #1860, opened 2026-09-03
[snrc resolver] implement resolving 2LD names by labelhash

brenzi:
Do not merge before deciding on merging #1861 first.

* support resolution by labelhash for 2LD name
* but not for subnames as there is no front-running risk for subname registration.
* add CI job


brenzi:
superseded by #1863, which now carries this work against master

PR #1861, opened 2026-09-03
[snrc resolver] report expiry and availability correctly

brenzi:
Problem with #1860 alone: It prepares the resolver to be used to query availability of names without revealing them, but the response can be misleading because we don't handle expiry and reserved status meaningfully. 
* An expired name beyond its grace period would still resolve although it is actually available again. 
* Reserved names would show as available although they aren't
* taken-but-resolverless name 404s identically to a free one

This PR fixes the problems above
* report expiry and reserved status (preparing for reporting a specific reason)
* introduce meaningful http error codes and handle them in `Names.hs`


brenzi:
superseded by #1863, which now carries this work against master

